PR #35, opened 2015-02-12
client: initial commit of go client

Sirupsen:
Starting to extract the helper methods from `api_test.go` into `client/` which can be used as a Go HTTP client.

I'll extract the toxics and fix the error messages in a follow up PR. Just wanted to commit tonight's work to commit to finishing the job. :-)

@wvanbergen this is primarily meant for you for `kafkabridge`. Please review along with @xthexder 

CI is failing due to subpackage thing... I'll take  a look tomorrow


xthexder:
Looks like a good start, should just need to fix the Godep problem and all is set.

PR #37, opened 2015-02-12
client: change to more idiomatic api

Sirupsen:
Changes to a more nice Go API where we can do things like:

```go
proxy.Create()
proxy.Enabled = false
proxy.Save()
proxy.Destroy()
client.Proxies()
```

Instead of doing everything on the client.

Follow up PRs will clean up errors, split into files, add toxic API.

@xthexder

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -13,18 +13,13 @@ import (
 
 var testServer *server
 
-var TestProxy = &tclient.Proxy{
+var client = tclient.NewClient("http://127.0.0.1:8475")
+var testProxy = client.NewProxy(&tclient.Proxy{
 	Name:     "mysql_master",
 	Listen:   "localhost:3310",
 	Upstream: "localhost:20001",
 	Enabled:  true,
-}
-
-type ProxyWithToxics struct {
-	Proxy
-	ToxicsUpstream   map[string]interface{} `json:"upstream_toxics"`
-	ToxicsDownstream map[string]interface{} `json:"downstream_toxics"`
-}
+})
 
 func WithServer(t *testing.T, f func(string)) {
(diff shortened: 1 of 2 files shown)


api_test.go
```diff
@@ -37,6 +27,12 @@ func WithServer(t *testing.T, f func(string)) {
 		time.Sleep(50 * time.Millisecond)
 	}
 
+	// Need to assign here since we need to call NewProxy to set the client.
+	testProxy.Name = "mysql_master"
+	testProxy.Listen = "localhost:3310"
+	testProxy.Upstream = "localhost:20001"
+	testProxy.Enabled = true
```
Sirupsen: This kind of sucks, but I'm not sure how to create a better interface unless we require passing a proxy object to `NewProxy`, which is also weird...

xthexder: Maybe pass in the name and upstream address to `NewProxy` and then the listen address defaults to :0 and enabled = true

Sirupsen: Perhaps, it's not really ideal either. I'll take a look at what other HTTP wrappers do.

xthexder:
Looks good :+1: will be really nice once it has the toxics added.

Sirupsen:
After talking with @boourns about how to do the `NewProxy` API I think the best compromise is that you pass a struct of your type to `NewProxy`:

```go
proxy = &Proxy{
  Listen = "localhost:38824",
  Upstream = "localhost:3306",
  Name = "mysql_something",
  Enabled = true
}

client.NewProxy(proxy)
```

What do yo uthink @xthexder? Then what `NewProxy` does is just set the `client` on it.

PR #44, opened 2015-04-12
client: add toxics and promote from wip

Sirupsen:
@eapache @xthexder

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -1,0 +1,10 @@
+# toxiproxy-go
+
+This is the Go client library for the
+[Toxiproxy](https://github.com/shopify/toxiproxy) API. Please read the [usage
+section in the Toxiproxy README](https://github.com/shopify/toxiproxy#usage)
+before attempting to use the client.
+
+For Usage please [see the Godoc
+documentation](http://godoc.org/github.com/Shopify/toxiproxy/client) for the
+package.
--- a/client/client.go
+++ b/client/client.go
@@ -57,7 +57,8 @@ func (client *Client) Proxies() (map[string]*Proxy, error) {
 }
 
 // NewProxy instantiates a new proxy instance. Note Create() must be called on
-// it to create it.
+// it to create it. The Enabled field must be set to true, otherwise the Proxy
+// will not be enabled when created.
 func (client *Client) NewProxy(proxy *Proxy) *Proxy {
 	if proxy == nil {
 		proxy = &Proxy{}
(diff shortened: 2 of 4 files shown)


client/README.md
```diff
@@ -0,0 +1,8 @@
+# toxiproxy-go
+
+This is the Go client library for the
+[Toxiproxy](https://github.com/shopify/toxiproxy) API. Please read the [usage
+section in the Toxiproxy README](https://github.com/shopify/toxiproxy#usage)
+before attempting to use the client.
+
+For Usage please see the Godoc documentation for the package.
```
xthexder: It woud be good to actually provide a link to the godocs.

client/client.go
```diff
@@ -46,6 +56,8 @@ func (client *Client) Proxies() (map[string]*Proxy, error) {
 	return proxies, nil
 }
 
+// NewProxy instantiates a new proxy instance. Note Create() must be called on
+// it to create it.
```
xthexder: This should probably note that the proxy is not enabled by default

xthexder:
Changes look good :+1: 

I noticed while reading through the main Readme again that the Curl example is out of date (it doesn't contain any of the proxy enabled/disabled stuff)
It would probably be a good idea to revisit that at some point.

Sirupsen:
Rebased and fixed your comments. Fixed the main README change in  https://github.com/Shopify/toxiproxy/commit/b979c77411aa6336d716a1f074985248a9e56cb2

client/client.go
```diff
@@ -1,3 +1,5 @@
+// The Toxiproxy Go client provides a wrapper around the Toxiproxy HTTP API for
```
eapache: By convention, package-level godoc starts with "Package $NAME"

client/client.go
```diff
@@ -151,6 +171,29 @@ func (proxy *Proxy) Toxics(direction string) (map[string]interface{}, error) {
 	return toxics, nil
 }
 
+// SetToxic sets the parameters for a toxic with a given name in the direction.
+// See https://github.com/Shopify/toxiproxy#toxics for a list of all Toxics.
+func (proxy *Proxy) SetToxic(name string, direction string, fields Fields) (map[string]interface{}, error) {
```
eapache: should direction (and maybe name) be enum types rather than arbitrary strings?

eapache:
A few minor comments, but :+1:

PR #47, opened 2015-04-28
Change latency toxic to account for buffering

xthexder:
@Sirupsen @pushrax 

Changes the latency toxic to better model network latency.

Previously the latency toxic would insert a delay after each "packet", and these would accumulate for large amounts of data.
This PR changes toxiproxy to calculate the delay based on the receive time of the "packet".

### DIFF
--- a/io_chan.go
+++ b/io_chan.go
@@ -5,23 +5,23 @@ import (
 	"time"
 )
 
-// Simulates a TCP packet by storing a slice of bytes with the receive timestmap
-type Packet struct {
+// Stores a slice of bytes with its receive timestmap
+type StreamChunk struct {
 	data      []byte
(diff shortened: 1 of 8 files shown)


pushrax:
All makes sense to me, but I think it's a bit confusing to call it a packet since there is no relation to actual packets. Just a naming nitpick though, otherwise it's fine.

Could be worth looking into #39 and implementing latency simulation at the TCP packet level or lower.

xthexder:
I'm not really sure what to call this other than a packet. Maybe just remove the reference to TCP in the comment?
Technically this is a packet of data, just not a TCP packet.

xthexder:
StreamChunk it is.

PR #48, opened 2015-04-28
Add Bandwidth toxic

xthexder:
@Sirupsen @pushrax 

Depends on https://github.com/Shopify/toxiproxy/pull/47

Adds a bandwidth toxic for limiting connections to a certain bitrate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.1.0 (Unreleased)
 
 * Remove /toxics endpoint in favour of /proxies
+* Add bandwidth toxic
 
 # 1.0.3
 
--- a/README.md
+++ b/README.md
@@ -292,6 +292,15 @@ Bringing a service down is not technically a toxic in the implementation of
 Toxiproxy. This is done by `POST`ing to `/proxies/{proxy}` and setting the
 `enabled` field to `false`.
 
+#### bandwidth
+
+Limit a connection to a maximum number of kilobytes per second.
+
+Fields:
+
+ - `enabled`: true/false
+ - `rate`: rate in KB/s
+
 #### slow_close
 
 Delay the TCP socket from closing until `delay` has elapsed.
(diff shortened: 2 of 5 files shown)


io_chan.go
```diff
@@ -1,20 +1,29 @@
 package main
 
-import "io"
+import (
+	"io"
+	"time"
+)
+
+// Simulates a TCP packet by storing a slice of bytes with the receive timestmap
```
Sirupsen: TCP packet doesn't really make sense, rather you're buffering a stream—but I suppose that's what you mean?

Sirupsen:
Testing this I can imagine is a little tough, but it shouldn't be impossible

xthexder:
Working on a go benchmark/test for this right now. I've tested this manually with curl already and it works well.

io_chan.go
xthexder: This is actually part of https://github.com/Shopify/toxiproxy/pull/47, just the second commit is for this PR.

pushrax: Comment still stands though.

pushrax:
Could write a test that sends sentinel data over the connection and checks every few quanta how much data has been received, not sure if it is necessary though.

xthexder:
The bandwidth toxic seems to be accurate from 1KB/s all the way up to 200MB/s (> 1gbit), after that it starts to get limited by the actual speed of the proxy.

Depending on the size of the writes going in, the noop benchmark produces some interesting numbers:
```
GOMAXPROCS=4
1kb writes: 70MB/s
10kb writes: 500MB/s
1mb writes: 680MB/s
10mb writes: 910MB/s
100mb writes: 800MB/s
```

I think the lower write sizes are actually being limited by the speed of the test suite, not the proxy.

xthexder:
Just added a latency test as well, I'm getting 100-150µs when I run it locally, which I think is pretty good.

Sirupsen:
Can you add the benchmark results to the FAQ and add this to the CHANGELOG? :)

xthexder:
I'm not really sure what to write in the FAQ anymore, Just ran the same benchmarks on my desktop at home and got 20µs latency and a crazy 2100MB/s peak (100mb writes)

Issue #36, opened 2015-02-12
Simplify by always using ProxyWithToxics


Sirupsen:
Is there any reason why the `Proxy` type doesn't just have `Toxics` always? It's perfectly fine for them to be nil. I also think proxies should be returned with toxics in all cases in the API. This simplifies things a great deal.

@xthexder thoughts?

xthexder:
The main reason was that listing toxics and proxies locks up certain parts of the code until the request finishes. Looking back at it, I don't think this actually affects the data flowing through the proxies, just other api calls.

I think it would be fine to always include the toxics in api responses. We can replace the `/proxies` endpoint with what is currently `/toxics`.

Sirupsen:
yeah, that's what I was thinking :+1:

PR #49, opened 2015-04-28
Remove /toxics endpoint in favour of /proxies

xthexder:
@Sirupsen 

Implements changes proposed in https://github.com/Shopify/toxiproxy/issues/36

Will have to look at what uses this endpoint. I think it may just be turbobar.
The ruby client does not use this endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.1.0 (Unreleased)
+
+* Remove /toxics endpoint in favour of /proxies
+
 # 1.0.3
 
 * Rename Go library package to Toxiproxy from Client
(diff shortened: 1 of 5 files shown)


Sirupsen:
Yes :heart_eyes: . Remember to remove it from the README and Go client. Also pinging @mdevilliers for .NET client.

eapache:
Also (although it is not specified explicitly anywhere) I believe you have been following semantic versioning, in which case this will force 2.0.

xthexder:
Maybe we can call this version 1.1
It's a pretty minor api change.

Looks like I forgot to update the CHANGELOG flle for 1.0.3 so I'll include that in this PR

pushrax:
@eapache is right if you do intend to follow semver.

Sirupsen:
I'm ok with breaking SemVer and doing 1.1

Issue #51, opened 2015-05-08
Return 400 when name is empty


Sirupsen:
When creating a new Proxy and name is empty it should return 400

PR #52, opened 2015-05-08
Change error handling, and check for blank proxy names

xthexder:
Fixes https://github.com/Shopify/toxiproxy/issues/51 and gives better errors back form the go api

@Sirupsen

### DIFF
--- a/api.go
+++ b/api.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"log"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


api.go
```diff
@@ -103,6 +104,15 @@ func (server *server) ProxyCreate(response http.ResponseWriter, request *http.Re
 		return
 	}
 
+	if len(input.Name) < 1 {
+		http.Error(response, server.apiError(errors.New("Missing required field: name"), http.StatusBadRequest), http.StatusBadRequest)
+		return
+	}
+	if len(input.Upstream) < 1 {
+		http.Error(response, server.apiError(errors.New("Missing required field: name"), http.StatusBadRequest), http.StatusBadRequest)
```
Sirupsen: name => upstream, we need listen as well perhaps?

api_test.go
```diff
@@ -61,6 +61,18 @@ func TestCreateProxy(t *testing.T) {
 	})
 }
 
+func TestCreateProxyBlankName(t *testing.T) {
```
Sirupsen: Test for upstream

Sirupsen:
oh wow thanks! I just left that issue there as a note to myself to fix it :heart:

api.go
xthexder: A blank listen field means use an ephemeral port, so that's allowed.

Sirupsen: Oh yes!

Sirupsen:
LGTM

Sirupsen:
bump :)

PR #53, opened 2015-05-09
PHP client

ihsw:
[Mostly complete and tested.](https://github.com/ihsw/toxiproxy-php-client)

In the near future I will be starting a [Symfony](http://symfony.com/) bundle as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ development and CI environments.
 * [toxiproxy-ruby](https://github.com/Shopify/toxiproxy-ruby)
 * [toxiproxy-go](https://github.com/Shopify/toxiproxy/tree/master/client)
 * [toxiproxy.net](https://github.com/mdevilliers/Toxiproxy.Net)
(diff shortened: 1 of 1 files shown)


xthexder:
I had a quick look through and it looks good.

Just a couple things:
In [examples/create.php](https://github.com/ihsw/toxiproxy-php-client/blob/master/examples/create.php) you're using an ephemeral port. You should only use ports < 32768 to prevent random conflicts. It's also possible to specify a blank `listen` field and have toxiproxy pick the port.

It would also be nice to have the client not enforce which toxics are available, since that's something that can change somewhat. I'll leave that up to you through if you want to keep it as-is.

pushrax:
@xthexder are you sure that any port < 32768 is safe? I think it depends on the kernel config, on linux check `/proc/sys/net/ipv4/ip_local_port_range`.

Sirupsen:
Hey, thanks for this! I'd like to see an example/documentation on how to use toxics as well :+1:

PR #55, opened 2015-05-13
*Tiny* changes to anchors in README

TkTech:
Anchors are case-sensitive by specification, so "#Clients" does not work.
Some anchors were not relative, and were corrected.


Sirupsen:
Thanks!

PR #57, opened 2015-05-18
client/go: proper types for toxics and collections

Sirupsen:
This makes using the client with toxics a lot more enjoyable!

@xthexder


xthexder:
:+1:

PR #58, opened 2015-05-18
cli: initial commit

Sirupsen:
Adds a simple V1 of a CLI that enables you to list proxies and toggle their enabled flag. Really nice to get an overview of things and play around with it. Will support Toxics later, and then we'll figure out how to distribute it. It could be a separate binary with a name like `toxiproxyctl` or it could be part of the main one and then daemon just becomes a command (and it starts by default to be backwards compatible).

This is really handy for quick and dirty testing yourself, or for black-box resiliency testing. A more friendly interface would be a web app, that can be done as well. It's remarkably easy with the Go client.

screen shot 2015-05-18 at 16 13 52
screen shot 2015-05-18 at 16 13 34
screen shot 2015-05-18 at 16 13 24

@xthexder

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,0 +1,132 @@
+package main
+
+import (
+	"github.com/Shopify/toxiproxy/client"
+	"github.com/codegangsta/cli"
+	"sort"
+
+	"fmt"
+	"os"
+)
+
+const (
+	redColor    = "\x1b[31m"
+	greenColor  = "\x1b[32m"
+	yellowColor = "\x1b[33m"
+	blueColor   = "\x1b[34m"
+	grayColor   = "\x1b[37m"
+	noColor     = "\x1b[0m"
+)
+
+func main() {
+	toxiproxyClient := toxiproxy.NewClient("http://localhost:8474")
+
+	app := cli.NewApp()
+	app.Name = "Toxiproxy"
+	app.Usage = "Simulate network and system conditions"
+	app.Commands = []cli.Command{
+		{
+			Name:    "list",
+			Aliases: []string{"l", "li", "ls"},
+			Usage:   "List all proxies",
+			Action: func(c *cli.Context) {
+				proxies, err := toxiproxyClient.Proxies()
+				if err != nil {
+					fmt.Println("Failed to retrieve proxies: ", err)
+					os.Exit(1)
+				}
+
+				var proxyNames []string
+				for proxyName := range proxies {
+					proxyNames = append(proxyNames, proxyName)
+				}
+				sort.Strings(proxyNames)
+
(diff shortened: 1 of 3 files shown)


xthexder:
:tada: amazing

I'd change the `ls` command to use more than just color to show up/down. Some terminals won't be able to see anything, and you can't really grep it either.

Sirupsen:
Great point :+1: I'll change that and ship.

Sirupsen:
What do you think about binary split vs in the same?

xthexder:
I don't really have much of a strong opinion one way or the other, though I think I'm leaning towards something like like `toxiproxy-cli` a la redis.

Sirupsen:
Then we should change the toxiproxy server component to `toxiproxy-server`. The nice thing with making it its own binary is that it's extremely discoverable.

ihsw:
:+1: to `toxiproxy-client` and `toxiproxy-server`

Sirupsen:
Just curious to hear some of the arguments for that design, @ihsw care to elaborate why you think it's better?

ihsw:
The nod to Redis is sensible, both Toxiproxy and Redis are small, fast, and easy to use.

I must acknowledge that the ctl suffix is not without precedent -- `apache2` and `apache2ctl`, `supervisor` and `supervisorctl`, and so forth -- and I've enjoyed working with them, but they're not as known for being lean and mean.

Sirupsen:
I added a `-server` and `-client` component and put it into the Debian package. These will now have to be included in releases, I'll release them as part of `1.2.0`. This is backwards compatible in terms of the binary names, but in `2.0.0` we should remove the server binary without the suffix.

cli/cli.go
```diff
@@ -0,0 +1,132 @@
...
+
+	app.Run(os.Args)
+}
+
+func enabledColor(enabled bool) string {
+	if enabled {
+		return "\x1b[32m"
+	}
+
+	return "\x1b[31m"
```
dwradcliffe: Should these use the constants defined above?

dwradcliffe:
This is really great! Can we get this shipped? @Sirupsen @xthexder

cli/cli.go
Sirupsen: Yeah, definitely... I've written these so many times I read them directly as the colours :blush:

jpittis:
Any chance I could help finish this up? What needs to be done besides color constants?
@Sirupsen

jpittis:
I thought I'd go ahead and make a PR [here](https://github.com/Shopify/toxiproxy/pull/91). :)

jpittis:
Closing this PR. Was rebased on to dev-2.0 here: #93

PR #59, opened 2015-05-22
docker: initial commit of dockerfile

Sirupsen:
This'll make it even easier to include it for development and CI in other projects that already use Docker and is extremely simple to do with Go.

@xthexder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Add a Toxic and Toxics type for the Go client
+* Add `Dockerfile`
 
 # 1.1.0
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM golang:1.4 
+
+ADD . /app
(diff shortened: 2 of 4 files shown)


toxiproxy.go
```diff
@@ -12,7 +12,7 @@ var host string
 var port string
 
 func init() {
-	flag.StringVar(&host, "host", "localhost", "Host for toxiproxy's API to listen on")
+	flag.StringVar(&host, "host", "0.0.0.0", "Host for toxiproxy's API to listen on")
```
Sirupsen: This is so it binds to all interfaces in Docker which makes sure that if you're using native networking it binds on the `docker0` interface so you can access it from other containers.

xthexder: This kindof seems wrong to me. Depending on where you're running it, you don't want just anyone to go and mess with your proxies.

Is there some way to just pass `-host 0.0.0.0` in the Dockerfile?

Sirupsen: Yeah, we can just do that in `CMD`.

xthexder:
:+1: :cake:

Issue #54, opened 2015-05-12
Toxiproxy 2.0 RFC


xthexder:
## Proposed API Changes
- Remove `enabled` field from all toxics
- `POST /proxies/{proxy}/{up|down}stream/toxics` - Create (enable) a new toxic
- `POST /proxies/{proxy}/{up|down}stream/toxics/{toxic}` - Update an existing toxic
- `DELETE /proxies/{proxy}/{up|down}stream/toxics/{toxic}` - Destroy (disable) a toxic

This API matches the current proxy creation/deletion api.

## Internal Structure Changes
Current 1.x architecture: https://docs.google.com/drawings/d/1pwuRINbpwyaQeYSgGt8VTEKWEUQAwYpNJIMH1owW3Sc/view

Instead of a static number of `ToxicStubs` (and a static mapping for each toxic), `ToxicStub`s will be dynamically added / removed as toxics are created/destroyed. A minimum of 1 `ToxicStub` will always be present, and the first toxic will always be a `NoopToxic` so that it can be interrupted for inserting / removing toxics.

**Steps for adding a new toxic:**
1. Interrupt the last toxic in the chain
2. Reconnect the output channels so that the new toxic is between the last toxic and the end `ChanReader`
3. Resume both `ToxicStub`s with their toxics

**Steps for removing a toxic:**
1. Interrupt the toxic before the target in order to ensure the target's input channel is empty
2. Interrupt the target toxic
3. Connect the previous toxic's output to the target toxic's output
4. Resume the previous toxic / `ToxicStub` pair

This ends up being very similar to a doubly-linked list, except that we need to ensure no data is in-flight between the links (channels) when they're removed.

## Toxic Registration
The `main()` function and server creation will be simplified so that it can be used as a library.
Instead of having all the toxics hard-coded inside toxiproxy, they will be registered in an `init()` function: `toxiproxy.Register(SomeToxic)` (this will have to be done with reflection or something...)

This will allow user-created toxics without having to create a full fork of toxiproxy.

pushrax:
What's the plan for ensuring the toxic has finished sending output after being interrupted, and before the connections are changed? Wait for `Pipe()` to return?

xthexder:
`stub.Interrupt()` will not return until `Pipe()` has returned now, so we are guaranteed that either the data is sent or it won't be until we resume.

pushrax:
:cool:

Sirupsen:
Wouldn't it be more restful to do:

[...]

PR #62, opened 2015-06-05
Toxiproxy 2.0

xthexder:
See proposal here https://github.com/Shopify/toxiproxy/issues/54

Ready for review + merging

TODO:
 - [x] Update Go client (and API tests)
 - [x] Add new tests for adding / removing / updating toxics
 - [x] Standardize error output in the API
 - [x] Run benchmark comparisons to v1.1
 - [x] Fix docker image builds
 - [ ] Add `toxicity` field: https://github.com/Shopify/toxiproxy/pull/65 (can be done in a separate PR after this)
 - [ ] Determine any deprecation / compatibility changes if necessary
 - [ ] Code cleanup and review

cc @Sirupsen @pushrax @eapache


xthexder:
Here's a quick summary of what has changed in this PR:

 - Toxics have moved into their own package, and are registered with an `init() {}` function.
 - Toxics no longer have an enabled field, all state such as name, index, etc... is stored in a wrapper now.
 - AddToxic / UpdateToxic / RemoveToxic have been added to the api and toxic_collection, as described in the RFC.
 - `io_chan.go` and related files have been moved to the `stream` package`
 - Many files have been moved to the `proxy` package

xthexder:
Alright, the package has been flattened again. The diff is a lot nicer to read now as well.
The code is definitely a lot cleaner now.

api.go
```diff
@@ -197,17 +226,25 @@ func (server *server) ProxyDelete(response http.ResponseWriter, request *http.Re
...
 
-	data, err := json.Marshal(proxyWithToxics(proxy))
+	var data []byte
+	switch vars["stream"] {
+	case "upstream":
+		data, err = json.Marshal(proxy.UpToxics.GetToxicMap())
+	case "downstream":
+		data, err = json.Marshal(proxy.DownToxics.GetToxicMap())
+	default:
+		http.Error(response, server.apiError(ErrInvalidStream, http.StatusBadRequest), http.StatusBadRequest)
```
eapache: missing a return

toxics/bandwidth_test.go
```diff
@@ -0,0 +1,119 @@
+package toxics_test
```
eapache: should this be `package toxics`?

version.go
```diff
@@ -0,0 +1,3 @@
+package toxiproxy
+
+var Version = "1.1.0"
```
eapache: should be 2.0?

eapache:
A few nits, but LGTM as much as I can really claim to understand the code

api.go
xthexder: Fixed

toxics/bandwidth_test.go
xthexder: All the tests are in a `_test` package now. I think this is prefered since we don't actually need to access internal state in these tests.

xthexder:
I've cleaned up the API response codes so they're consistent.
`version.go` is also auto-updated via the `VERSION` file when `make` is run

I don't think any extra backwards compatibility stuff will be realistic to add. 1.2.1 is stable, so it should be fine for people to use that untl all the clients are updated.
The `toxicity` field will probably be pushed back to a separate PR, since this one is getting big enough as it is.

Sirupsen:
hey @xthexder, I need to review this. I'm sorry I haven't gotten to it.

Sirupsen:
Is the easiest way to do the individual commits?

xthexder:
I think the easiest way is just to just go through the final diff. I tried to make the individual commits readable, but it kindof failed.
There's very few file moves now, so the final diff isn't as messy as it used to be.

xthexder:
Just did some :tophat:-ing on this. Found a couple things that needed to be fixed with the api endpoints.
The actual proxy portion of this looks stable :+1:

xthexder:
I just completed a massive rebase to make reviewing easier.
The first commit is the guts of everything in terms of 2.0 features.
The second commit cleans up api resposne codes and error handling in the api.
The last commit is a noop in terms of code, and just contains a bunch of tile renames and package references.

xthexder:
I've been going through this code, doing a bunch of thorough testing, and found another problem:

Due to https://github.com/Shopify/toxiproxy/pull/67 the logic in the RFC isn't completely correct. `InterruptToxic()` doesn't empty the buffer anymore (and shouldn't for other reasons), so when we remove a toxic it's possible to drop data.
This is closely related to https://github.com/Shopify/toxiproxy/issues/72, so I'll try to solve both issues when I fix this.

xthexder:
The latest commit should fix the issues I mentioned above. Toxics can specify how large a buffer they need (currently only the latency toxic needs one), and it will be handled properly.
I also added some more tests to make sure data isn't dropped.

Once again, I would consider this ready for review.

api.go
```diff
@@ -29,10 +30,11 @@ func (server *server) Listen(host string, port string) {
...
 	r.HandleFunc("/proxies/{proxy}", server.ProxyDelete).Methods("DELETE")
-	r.HandleFunc("/proxies/{proxy}/upstream/toxics", server.ToxicIndexUpstream).Methods("GET")
-	r.HandleFunc("/proxies/{proxy}/downstream/toxics", server.ToxicIndexDownstream).Methods("GET")
-	r.HandleFunc("/proxies/{proxy}/upstream/toxics/{toxic}", server.ToxicSetUpstream).Methods("POST")
-	r.HandleFunc("/proxies/{proxy}/downstream/toxics/{toxic}", server.ToxicSetDownstream).Methods("POST")
+	r.HandleFunc("/proxies/{proxy}/{stream}/toxics", server.ToxicIndex).Methods("GET")
+	r.HandleFunc("/proxies/{proxy}/{stream}/toxics", server.ToxicCreate).Methods("POST")
+	r.HandleFunc("/proxies/{proxy}/{stream}/toxics/{toxic}", server.ToxicShow).Methods("GET")
+	r.HandleFunc("/proxies/{proxy}/{stream}/toxics/{toxic}", server.ToxicUpdate).Methods("POST")
+	r.HandleFunc("/proxies/{proxy}/{stream}/toxics/{toxic}", server.ToxicDelete).Methods("DELETE")
```
Sirupsen: Should we consider removing the `stream` from the REST URL and making it simply a property of the toxic?

link_test.go
```diff
@@ -0,0 +1,148 @@
+package toxiproxy
```
Sirupsen: What happens if you add multiple of the same toxic?

api.go
xthexder: Hmmm, that could clean things up a bit, though it would require a couple internal changed, since up/downstream toxics can have the same name.
We could default to downstream though, and it would simplify the usage.

testing/benchmark_test.go
```diff
@@ -8,17 +8,27 @@ import (
...
+//     BenchmarkDirectSmall        5000            291324 ns/op
+//     BenchmarkProxySmall         3000            504501 ns/op
+//
+// 10x Toxic Types:
+//     BenchmarkDirect             3000            599519 ns/op
+//     BenchmarkProxy              2000           1044746 ns/op
+//     BenchmarkDirectSmall        5000            280713 ns/op
+//     BenchmarkProxySmall         3000            574816 ns/op
+//
+// Toxiproxy 2.0
```
Sirupsen: What about a benchmark with enabled toxics?

toxic_collection.go
```diff
@@ -43,58 +38,137 @@ func (c *ToxicCollection) ResetToxics() {
...
+
+	result := make(map[string]toxics.Toxic)
+	for _, toxic := range c.toxics {
+		result[toxic.Name] = toxic
 	}
 	return result
 }
 
-func (c *ToxicCollection) SetToxicJson(name string, data io.Reader) (Toxic, error) {
+func (c *ToxicCollection) AddToxicJson(data io.Reader) (toxics.Toxic, error) {
```
Sirupsen: Why is this in here? This sounds like a responsibility of the API rather than the collection.

toxic_collection.go
```diff
@@ -43,58 +38,137 @@ func (c *ToxicCollection) ResetToxics() {
...
+	if err != nil {
+		return nil, joinError(err, ErrBadRequestBody)
+	}
+
+	c.toxics = append(c.toxics, wrapper)
+	c.chainAddToxic(wrapper)
+	return wrapper.Toxic, nil
+}
+
+func (c *ToxicCollection) UpdateToxicJson(name string, data io.Reader) (toxics.Toxic, error) {
```
Sirupsen: Ditto

api.go
Sirupsen: I like that.

Sirupsen:
Can we add changelog entries as well?

api.go
Sirupsen: You could merge this into a 2.0 branch, and then play with that.

link_test.go
xthexder: Yep, you can add multiple of the same type, as long as they have a different name.
Tested here: https://github.com/Shopify/toxiproxy/pull/62/files#diff-095e965c1e92c9277036371d89bd083cR409

Sirupsen: :+1:

testing/benchmark_test.go
xthexder: These are actually benchmarks with no toxics enabled, but N toxics loaded / available.
The number of available toxics is no longer a performance factor in 2.0 since we don't run data through N noop's anymore.
There are some benchmark tests for specific toxics, but I think the closest comparison would be enabling N noop toxics in 2.0

toxic_collection.go
xthexder: It's in here because it needs the lock on ToxicCollection (and because it was in here before...)
The json decoding is a little messy since it needs to figure out what type of toxic it is before it can do the decode (requires access to ToxicCollection)

Sirupsen:
Nothing jumps out, but there's a lot of code making it tough to do a thorough review. Really excited about this :+1:

xthexder:
There's effectively a changelog in the PR comments, but I'll add the actually file to the PR.

Sirupsen:
Yep, that's what I meant :)

xthexder:
Merged to the [2.0-dev branch](https://github.com/Shopify/toxiproxy/tree/2.0-dev), will continue development in separate PRs.

PR #63, opened 2015-07-02
Node Client

dlion:
I wrote a node client.
You can find it here: http://github.com/DLion/toxiproxy-node


Sirupsen:
Amazing! Thank you :)

PR #65, opened 2015-07-14
Add `sometimesToxic` and `toxicity` options to `timeout`

tomblench:
This PR is intended to create discussion about whether people think introducing randomness into Toxiproxy is a good idea.

At our organisation we have just started using Toxiproxy and really like it, but we would like to add random failures.

To introduce randomness to the `timeout` Toxic, set `sometimesToxic` in the json request to `true`, and set the required `toxicity`. The reason for using two variables is to maintain compatibility with existing behaviour - if you don't explicitly set `sometimesToxic` then it will default to `false` and 100% of requests will timeout.

It's also possible that this could be generically extended to all Toxics - it looks like the work on the `toxiproxy2` branch could make this easier.

### DIFF
--- a/toxic_timeout.go
+++ b/toxic_timeout.go
@@ -1,6 +1,9 @@
 package main
 
 import "time"
+import "math/rand"
+
+var noop = new(NoopToxic)
 
 // The TimeoutToxic stops any data from flowing through, and will close the connection after a timeout.
 // If the timeout is set to 0, then the connection will not be closed.
@@ -8,6 +11,10 @@ type TimeoutToxic struct {
 	Enabled bool `json:"enabled"`
 	// Times in milliseconds
 	Timeout int64 `json:"timeout"`
+	// If true, use Toxicity
+	SometimesToxic bool `json:"sometimesToxic"`
+	// 'Toxicity' or probability of timing out 0..1
+	Toxicity float32 `json:"toxicity"`
 }
 
 func (t *TimeoutToxic) Name() string {
@@ -23,17 +30,23 @@ func (t *TimeoutToxic) SetEnabled(enabled bool) {
 }
 
 func (t *TimeoutToxic) Pipe(stub *ToxicStub) {
-	timeout := time.Duration(t.Timeout) * time.Millisecond
-	if timeout > 0 {
-		select {
-		case <-time.After(timeout):
-			stub.Close()
-			return
-		case <-stub.interrupt:
+	if t.SometimesToxic && rand.Float32() >= t.Toxicity {
+		// just pipe the data through
+		noop.Pipe(stub)
+	} else {
+		// do timeout
+		timeout := time.Duration(t.Timeout) * time.Millisecond
+		if timeout > 0 {
+			select {
+			case <-time.After(timeout):
+				stub.Close()
+				return
+			case <-stub.interrupt:
+				return
+			}
+		} else {
+			<-stub.interrupt
 			return
 		}
-	} else {
-		<-stub.interrupt
-		return
 	}
 }


xthexder:
I definitely like the idea of a "affect n% of connections" option, and it certainly makes the timeout toxic more useful.
Having a global `toxicity` option that just defaults to 1 would make this cleaner, but that isn't too easy to implement without the `toxiproxy2` changes.

Randomized testing is certainly something I've kept in mind and is something I'd like to support more of in 2.0

tomblench:
Thanks for the feedback, I agree a default of 1 would be cleaner.

Is it worth me submitting a PR to `toxiproxy2` or should I hold off until `toxiproxy2` merges to master?

xthexder:
I expect the `toxiproxy2` to have a few more breaking changes happen to it before it's merged, so I'd probably hold off on a PR.
I'll add a reference to this in the 2.0 PR so it gets looked at.

Sirupsen:
We already support randomness in the `latency` toxic via the `jitter` parameter, and the `slicer` toxic in #66 also supports it. As long as it's optional, I'm perfectly happy with including it.

toxic_timeout.go
```diff
@@ -8,6 +11,10 @@ type TimeoutToxic struct {
 	Enabled bool `json:"enabled"`
 	// Times in milliseconds
 	Timeout int64 `json:"timeout"`
+	// If true, use Toxicity
+	SometimesToxic bool `json:"sometimesToxic"`
+	// 'Toxicity' or probability of timing out 0..1
+	Toxicity float32 `json:"toxicity"`
```
Sirupsen: Why have both `SometimesToxic` and `Toxicity`? Toxicity could just default to `1` (0 doesn't make sense).

Sirupsen:
This would need a changelog + README (updating the docs for the toxic) as well.

toxic_timeout.go
xthexder: I think this was just because of compatibility. There isn't a way to specify a default value without changing the toxiproxy core.

Sirupsen: The default could be implemented in `Pipe`.

Sirupsen:
I think `toxicity` should still be per toxic, not global. E.g. you may want latency more frequently than timeouts. I also don't want to burden everyone with implementing `toxicity` support for every toxic, it may not make sense for some.

xthexder:
The way I see this working, is that it would be a per-toxic setting, but implemented into the toxiproxy core. This way no toxics have to do anything special to get this functionality.

The logic would be very similar to what is done here: Add the real toxic on X% of the links, add a noop toxic to the rest.

Sirupsen:
Ah yeah I see what you mean. I think that's pretty reasonable. What would the API look like then?

xthexder:
`toxicity` is field we could add the to toxic wrapper: https://github.com/Shopify/toxiproxy/blob/toxiproxy2/toxics/toxic.go#L28
I think this will be pretty simple to implement with the 2.0 refactor. Not so much on 1.1 though.

You'd be posting something like this to create the toxic: `{"type": "timeout", "toxicity": 0.5, "timeout": 5000}`

Sirupsen:
That sounds like a good idea :+1: I also like that name.

toxic_timeout.go
xthexder: It's a bit of a hack, but it's less hacky than 2 fields I guess. (No way to have a 0 toxicity toxic, but why would you want that anyway?)

Sirupsen: Yeah exactly. Maybe that's an OK compromise for Toxiproxy 1.x before we have our own `toxicity` field to be forwards compatible. Then we can simply remove this logic in 2.x and it'll be backwards compatible.

xthexder:
Closed via https://github.com/Shopify/toxiproxy/pull/75
This will be available once 2.0 is released.

Sirupsen:
@tomblench this has been released as part of 2.0!

PR #66, opened 2015-07-16
Add toxic slicer

connor4312:
When working on relative low-level network applications/clients, it's easy to assume that TCP data will always some in nice grouped "packets" that have all the data you need in them, rather than viewing the connection as a stream. When working locally that may even be its apparent behaviour, but over real-world network connections that's certainly not something that can be relied upon.

This PR adds a "slicer" filter which takes data and slices it up into nice little random chunks before sending them out. The slider also includes a simple delay option, to bypass buffering behaviour that could otherwise nullify the sliced bytes.

### DIFF
--- a/toxic_slicer.go
+++ b/toxic_slicer.go
@@ -10,10 +10,10 @@ import (
 type SlicerToxic struct {
 	Enabled bool `json:"enabled"`
 	// Average number of bytes to slice at
-	AverageSize int `json:"averageSize"`
+	AverageSize int `json:"average_size"`
 	// +/- bytes to vary sliced amounts. Must be less than
 	// the average size
-	SizeVariation int `json:"sizeVariation"`
+	SizeVariation int `json:"size_variation"`
 	// Microseconds to delay each packet. May be useful since there's
 	// usually some kind of buffering of network data
 	Delay int `json:"delay"`
@@ -54,9 +54,9 @@ func (t *SlicerToxic) chunk(start int, end int) []int {
 	left := t.chunk(start, mid)
 	right := t.chunk(mid, end)
 
-	out := make([]int, len(left)+len(right))
-	copy(out, left)
-	copy(out[len(left):], right)
+	out := make([]int, 0, len(left)+len(right))
+	out = append(out, left...)
+	out = append(out, right...)
 
 	return out
 }
@@ -73,9 +73,9 @@ func (t *SlicerToxic) Pipe(stub *ToxicStub) {
 			}
 
 			chunks := t.chunk(0, len(c.data))
-			for i := 0; i < len(chunks); i += 2 {
+			for i := 1; i < len(chunks); i += 2 {
 				stub.output <- &StreamChunk{
-					data:      c.data[chunks[i]:chunks[i+1]],
+					data:      c.data[chunks[i-1]:chunks[i]],
 					timestamp: c.timestamp,
 				}
 
--- a/toxic_slicer.go
+++ b/toxic_slicer.go
@@ -81,6 +81,10 @@ func (t *SlicerToxic) Pipe(stub *ToxicStub) {
 
 				select {
 				case <-stub.interrupt:
+					stub.output <- &StreamChunk{
+						data:      c.data[chunks[i]:],
+						timestamp: c.timestamp,
+					}
 					return
 				case <-time.After(time.Duration(t.Delay) * time.Microsecond):
 				}
--- a/toxic_slicer.go
+++ b/toxic_slicer.go
@@ -50,7 +50,7 @@ func (t *SlicerToxic) chunk(start int, end int) []int {
 
 	// +1 in the size variation to offset favoring of smaller
 	// numbers by integer division
-	mid := start + (end-start)/2 + (rand.Intn(t.SizeVariation*2+1) - t.SizeVariation)
+	mid := start + (end-start)/2 + (rand.Intn(t.SizeVariation*2) - t.SizeVariation) + rand.Intn(1)
 	left := t.chunk(start, mid)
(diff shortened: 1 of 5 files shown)


pushrax:
Great idea for a toxic!

xthexder:
This is actually really cool. I found a bug in one of my apps about a week ago related to this exact problem.
As it is, toxiproxy will cause this behaviour in some cases (usually if latency or bandwidth toxics are enabled), but having it explicit like this is great.

I'm not the biggest fan of adding the delay in this toxic, since it's semi-duplicated functionality and reduces the bandwidth the toxic can process.
I haven't looked at this in-depth but [TCPConn.SetWriteBuffer](http://golang.org/pkg/net/#TCPConn.SetWriteBuffer) might be another option.

toxic_slicer.go
```diff
@@ -0,0 +1,90 @@
...
+	"math/rand"
+	"time"
+)
+
+// The SlicerToxic slices data into multiple smaller packets
+// to simulate real-world TCP behaviour.
+type SlicerToxic struct {
+	Enabled bool `json:"enabled"`
+	// Average number of bytes to slice at
+	AverageSize int `json:"averageSize"`
```
xthexder: The convention we've been using for json would be `average_size`.

toxic_slicer.go
```diff
@@ -0,0 +1,90 @@
...
+		return []int{start, end}
+	}
+
+	// +1 in the size variation to offset favoring of smaller
+	// numbers by integer division
+	mid := start + (end-start)/2 + (rand.Intn(t.SizeVariation*2+1) - t.SizeVariation)
+	left := t.chunk(start, mid)
+	right := t.chunk(mid, end)
+
+	out := make([]int, len(left)+len(right))
```
xthexder: This could be simplified to just `return append(left, right...)`

toxic_slicer.go
```diff
@@ -0,0 +1,90 @@
...
+		case <-stub.interrupt:
+			return
+		case c := <-stub.input:
+			if c == nil {
+				stub.Close()
+				return
+			}
+
+			chunks := t.chunk(0, len(c.data))
+			for i := 0; i < len(chunks); i += 2 {
```
xthexder: This nit-pick really doesn't matter, but I would start at `i = 1` and use `i-1, i`.
`len(chunks)` will always be even, but if for some odd reason it wasn't, this would prevent a panic.

connor4312:
I've adjusted it in response to the comments.

As far as the delay goes: I agree it's not a beautiful solution, but often the OS and/or standard networking libraries buffer input until it's ready to be processed. Go sets TCP NoDelay to be true by default, so there should be no outgoing buffer, but the consumer's setup is unknown. This was the simplest way I saw around that; the other was to change the order so that the latency filter was run after the slicer, but requiring two toxics to be used in tandem to be effective in many cases seemed awkward. If there's a more graceful solution I'm all for it.

toxic_slicer.go
```diff
@@ -0,0 +1,90 @@
...
+			chunks := t.chunk(0, len(c.data))
+			for i := 1; i < len(chunks); i += 2 {
+				stub.output <- &StreamChunk{
+					data:      c.data[chunks[i-1]:chunks[i]],
+					timestamp: c.timestamp,
+				}
+
+				select {
+				case <-stub.interrupt:
+					return
```
xthexder: If this toxic is interrupted at this point in the loop, the remainder of `c` will be dropped. In this case we would want to just send off the remainder like in the latency toxic.
Toxics can be interrupted without the stream closing, so we need to make sure it's in a usable state.

connor4312: Fixed

toxic_slicer.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// +1 in the size variation to offset favoring of smaller
+	// numbers by integer division
+	mid := start + (end-start)/2 + (rand.Intn(t.SizeVariation*2) - t.SizeVariation) + rand.Intn(1)
+	left := t.chunk(start, mid)
+	right := t.chunk(mid, end)
+
+	out := make([]int, 0, len(left)+len(right))
+	out = append(out, left...)
+	out = append(out, right...)
```
xthexder: The reason I suggested `append()` vs `copy()` is because this could be done in 1 line instead of 4. `append()` handles all the allocation and copying for you.

xthexder:
I think that's it for my comments.
@Sirupsen maybe you want to take a look at this too.

toxic_slicer.go
connor4312: It does, but doing so needs multiple memory allocations behind-the-scenes, and it allocates a greedy amount of memory (`2^x`). In this case we're lucky to "know" what information we want to be in the slice, so it's faster just to allocate exactly how much we need in a single go.

I might be a little OCD about performance :smile:

xthexder: Just did a quick benchmark since you got me curious.
The memory allocation is the same unless `len(right) < len(left)`, in which case using `make()` will use a bit less.
In almost all cases the size of `left` and `right` will be almost exactly the same (depending on `SizeVariation`), so I'm going to say I'd prefer shorter code of marginally faster code.

If performance is an issue with this, then the bigger target would be to not do recursive heap allocations.

connor4312: Fair enough, I've adjusted it :)

Sirupsen:
@connor4312 LGTM, nice job! can you add it to the CHANGELOG as well?

connor4312:
Thanks! Changelog has been updated.

connor4312:
Rebased off master due to conflict

Sirupsen:
@connor4312 released as part of 1.2.0

PR #67, opened 2015-07-20
Fix latency bandwidth

xthexder:
Due to the blocking channels inbetween toxics, the latency toxic was causing TCP reads to block, and thus slowing down the the maximum bandwidth of the toxic.
This change allows some buffering between toxics, which fixes the issue (until the buffer fills up, which will only happen at extremely high throughput + high latency configurations)

@Sirupsen @pushrax for review


pushrax:
Assuming that test fails without the changed code, :+1:

PR #70, opened 2015-08-03
Added a windows build to the makefile

mdevilliers:
It would be really useful for me to be able to link to an official windows executable for the [.net client](https://github.com/mdevilliers/toxiproxy).

To aid this I've added a build to the makefile to produce an windows exe for amd64.

To enable cross compilation of windows on linux, darwin follow the instructions at https://github.com/golang/go/wiki/WindowsCrossCompiling. 

then run make.bash enabling windows and amd64

```
cd $GOROOT/src
GOOS=windows GOARCH=amd64 ./make.bash --no-clean
```

Mark

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,12 +5,13 @@ GODEP_PATH=$(shell pwd)/Godeps/_workspace
 ORIGINAL_PATH=$(shell echo $(GOPATH))
 COMBINED_GOPATH=$(GODEP_PATH):$(ORIGINAL_PATH)
 
-.PHONY: packages deb test linux darwin
+.PHONY: packages deb test linux darwin windows
 
-all: deb linux darwin docker
+all: deb linux darwin windows docker
 deb: $(DEB)
 darwin: tmp/build/toxiproxy-darwin-amd64 
 linux: tmp/build/toxiproxy-linux-amd64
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,12 +5,13 @@ GODEP_PATH=$(shell pwd)/Godeps/_workspace
 ORIGINAL_PATH=$(shell echo $(GOPATH))
 COMBINED_GOPATH=$(GODEP_PATH):$(ORIGINAL_PATH)
 
-.PHONY: packages deb test linux darwin
+.PHONY: packages deb test linux darwin windows
 
-all: deb linux darwin docker
+all: deb linux darwin docker windows
```
xthexder: I would put `windows` before docker here, just so it builds all 3 binaries together.

xthexder:
:+1: I just tested toxiproxy on a windows machine, and it looks like it runs fine, though I noticed there's no support for ANSI color codes by default, so the logs have extra characters all over them (something I may have a look at later).

I've added a windows binary to the toxiproxy 1.2.1 release. If you'd like you could write something for the `Installing` section in the README.

mdevilliers:
Hey,

I've added a section on installing on Windows and altered the order of the builds as suggested.

Mark

PR #71, opened 2015-08-04
Fix slicer test race / deadlock

xthexder:
@Sirupsen @pushrax 

If the timeout is hit, the tests will hang on `output <- data` inside the toxic, this fixes that.
I also increased the timeout slightly since I ran into it once when my VM was running a little slow.

cc @connor4312


connor4312:
:+1:

pushrax:
:shipit:

PR #73, opened 2015-08-21
Remove stream direction from api urls

xthexder:
@Sirupsen @eapache 

As suggested in PR https://github.com/Shopify/toxiproxy/pull/62


Sirupsen:
Don't see anything wrong with the code, rad.

Can you update the documentation since this field is now required?

xthexder:
I set the default for this field to `downstream` since in most cases that's the most useful (testing services being slow), docs have been started in https://github.com/Shopify/toxiproxy/pull/74

Sirupsen:
cool

PR #74, opened 2015-08-21
Update README for 2.0

xthexder:
- [x] Update api docs
 - [x] Add a "Upgrading from 1.x" section
 - [x] Add a "Writing custom toxics" section
 - [x] Document Go client https://github.com/Shopify/toxiproxy/pull/76
 - [x] Add `toxicity` field docs https://github.com/Shopify/toxiproxy/pull/75

### DIFF
--- a/CREATING_TOXICS.md
+++ b/CREATING_TOXICS.md
@@ -1,0 +1,137 @@
+# Creating custom toxics
+
+Creating a toxic is done by implementing the `Toxic` interface:
+
+```go
+type Toxic interface {
+    Pipe(*toxics.ToxicStub)
+}
+```
+
+The `Pipe()` function defines how data flows through the toxic, and is passed a
+`ToxicStub` to operate on. A `ToxicStub` stores the input and output channels for
+the toxic, as well as an interrupt channel that is used to pause operation of the
+toxic.
+
+The input and output channels in a `ToxicStub` send and receive `StreamChunk` structs,
+which are similar to network packets. A `StreamChunk` contains a `byte[]` of stream
+data, and a timestamp of when Toxiproxy received the data from the client or server.
+
+Toxics are registered in an `init()` function so that they can be used by the server:
+```go
+func init() {
+    toxics.Register("toxic_name", new(ExampleToxic))
+}
+```
+
+## A basic toxic
+
+The most basic implementation of a toxic is the [noop toxic](https://github.com/Shopify/toxiproxy/blob/master/toxics/noop.go),
+which just passes data through without any modifications.
+
+```go
+type NoopToxic struct{}
+
+func (t *NoopToxic) Pipe(stub *toxics.ToxicStub) {
+    for {
+        select {
+        case <-stub.Interrupt:
+            return
+        case c := <-stub.Input:
+            if c == nil {
+                stub.Close()
+                return
+            }
+            stub.Output <- c
+        }
+    }
+}
+```
+
+The above code reads from `stub.Input` in a loop, and passes the `StreamChunk` along to
+`stub.Output`. Since reading from `stub.Input` will block until a chunk is available,
+we need to check for interrupts as the same time.
+
+Toxics will be interrupted whenever they are being updated, or possibly removed. This can
+happen at any point within the `Pipe()` function, so all blocking operations (including sleep),
+should be interruptible. When an interrupt is received, the toxic should return from the `Pipe()`
+function after it has written any "in-flight" data back to `stub.Output`. It is important that
+all data read from `stub.Input` is passed along to `stub.Output`, otherwise the stream will be
+missing bytes and become corrupted.
+
+When an `end of stream` is reached, `stub.Input` will begin returning `nil` chunks. Whenever a
+nil chunk is returned, the toxic should call `Close()` on the stub, and return.
+
+## Toxic configuration
+
+Toxic configuration information can be stored in the toxic struct. The toxic will be json
+encoded and decoded by the api, so all public fields will be api accessible.
+
+An example of a toxic that uses configuration values is the [latency toxic](https://github.com/Shopify/toxiproxy/blob/master/toxics/latency.go)
+
+```go
+type LatencyToxic struct {
+    Latency int64 `json:"latency"`
+    Jitter  int64 `json:"jitter"`
+}
+```
+
+These fields can be used inside the `Pipe()` function, but generally should not be written
+to from the toxic. A separate instance of the toxic exists for each connection through the
+proxy, and may be replaced when updated by the api. If state is required in your toxic, it
+is better to use a local variable at the top of `Pipe()`, since struct fields are not
+guaranteed to be persisted across interrupts.
+
+## Toxic buffering
+
+By default, toxics are not buffered. This means that writes to `stub.Output` will block until
+either the endpoint or another toxic reads it. Since toxics are chained together, this means
+not reading from `stub.Input` will block other toxics (and endpoint writes) from operating.
+If this is not behavior you want your toxic to have, you can specify a buffer size for your
+toxic's input. The [latency toxic](https://github.com/Shopify/toxiproxy/blob/master/toxics/latency.go)
+uses this in order to prevent added latency from limiting the proxy bandwidth.
+
+Specifying a buffer size is done by implementing the `BufferedToxic` interface, which adds the
+`GetBufferSize()` function:
+
+```go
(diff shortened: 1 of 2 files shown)


xthexder:
@Sirupsen @eapache 

As far as I know, this is the last thing before 2.0 can be released. It would be good if you guys could read through the docs :+1:

README.md
```diff
@@ -374,21 +394,35 @@ Changing the `listen` or `upstream` fields will restart the proxy and drop any a
...
+ - **GET /proxies/{proxy}** - Show the proxy with all its active toxics
  - **POST /proxies/{proxy}** - Update a proxy's fields
  - **DELETE /proxies/{proxy}** - Delete an existing proxy
- - **GET /proxies/{proxy}/upstream/toxics** - List upstream toxics
- - **GET /proxies/{proxy}/downstream/toxics** - List downstream toxics
- - **POST /proxies/{proxy}/upstream/toxics/{toxic}** - Update upstream toxic
- - **POST /proxies/{proxy}/downstream/toxics/{toxic}** - Update downstream toxic
- - **GET /reset** - Enable all proxies and disable all toxics
+ - **GET /proxies/{proxy}/toxics** - List active toxics
+ - **POST /proxies/{proxy}/toxics/{toxic}** - Update an active toxic
```
eapache: is there a `GET /proxies/{proxy}/toxics/{toxic}`?

README.md
```diff
@@ -374,21 +394,35 @@ Changing the `listen` or `upstream` fields will restart the proxy and drop any a
...
+ - **DELETE /proxies/{proxy}/toxics/{toxic}** - Remove an active toxic
+ - **GET /reset** - Enable all proxies and remove all active toxics
```
eapache: shouldn't this be a POST?

eapache:
a few minor questions but generally LGTM

README.md
xthexder: Whoops, yes there is.

CREATING_TOXICS.md
```diff
@@ -0,0 +1,137 @@
...
+                stub.Close()
+                return
+            }
+            stub.Output <- c
+        }
+    }
+}
+```
+
+The above code reads from `stub.Input` in a loop, and passes the `StreamChunk` along to
```
Sirupsen: You should briefly describe what a `StreamChunk` is; i.e. why is it not just `[]byte`.

CREATING_TOXICS.md
```diff
@@ -0,0 +1,137 @@
...
+`stub.Output`. Since reading from `stub.Input` will block until a chunk is available,
+we need to check for interrupts as the same time.
+
+Toxics will be interrupted whenever they are being updated, or possibly removed. This can
```
Sirupsen: Can you give an example of where this hookpoint is useful? What would a Toxic want to clean up?

CREATING_TOXICS.md
```diff
@@ -0,0 +1,137 @@
...
+happen at any point within the `Pipe()` function, so all blocking operations (including sleep),
+should be interruptible. When an interrupt is received, the toxic should return from the `Pipe()`
+function after it has written any "in-flight" data back to `stub.Output`. It is important that
+all data read from `stub.Input` is passed along to `stub.Output`, otherwise the stream will be
+missing bytes and become corrupted.
+
+When an `end of stream` is reached, `stub.Input` will begin returning `nil` chunks. Whenever a
```
Sirupsen: Why would it return more than one?

CREATING_TOXICS.md
```diff
@@ -0,0 +1,137 @@
+# Creating custom toxics
+
+Creating a toxic is done by implementing the `Toxic` interface:
```
Sirupsen: Add something here about the fact they have to be merged (or maintained) in a fork currently, but for the most part the toxics built in are all we use, so we don't think this is a huge concern. You can give some suggestions of protocol-specific toxics, e.g. a Redis toxic where you can drop percentages of reads and writes. However, with a note that it may not be terribly useful but for edge-cases as your worst case are still those provided by the latency toxic and downtime.

README.md
```diff
@@ -289,11 +293,21 @@ end
 
 Please consult your respective client library on usage.
 
+### Upgrading from Toxiproxy 1.x
```
Sirupsen: Perhaps put this under the installation step as a sub-section?

CREATING_TOXICS.md
Sirupsen: Ah, you already wrote a bit about this in the README. You should echo it here.

README.md
```diff
@@ -347,11 +357,21 @@ sliced "packet".
...
  - `average_size`: size in bytes of an average packet
  - `size_variation`: variation in bytes of an average packet (should be smaller than average_size)
  - `delay`: time in microseconds to delay each packet by
 
+### Writing custom toxics
+
+Writing your own toxics for Toxiproxy can be done by copying [toxiproxy.go](https://github.com/Shopify/toxiproxy/blob/master/cmd/toxiproxy.go)
+into a new project and registering your toxic with the server. This will allow you to build your own Toxiproxy
+binary without having to make a full fork of the project. If you think your toxics will be useful to others,
+contribute them back with a Pull Request.
```
Sirupsen: Do we really think people are going to import the project and build their own servers though?

Sirupsen: I still think it should be documented it's possible, but perhaps not in the main README.

README.md
```diff
@@ -374,21 +394,35 @@ Changing the `listen` or `upstream` fields will restart the proxy and drop any a
 If `listen` is specified with a port of 0, toxiproxy will pick an ephemeral port. The `listen` field
 in the response will be updated with the actual port.
 
-If you change `enabled` to `false`, it'll take down the proxy. You can switch it
+If you change `enabled` to `false`, it will take down the proxy. You can switch it
 back to `true` to reenable it.
 
+#### Toxic fields:
```
Sirupsen: Clarify that these are the fields on all toxics (global) and that a toxic may extend the struct with additional.

README.md
```diff
@@ -374,21 +394,35 @@ Changing the `listen` or `upstream` fields will restart the proxy and drop any a
...
+
+ - `name`: toxic name (string, defaults to `type`)
+ - `type`: toxic type (string)
+ - `stream`: link direction to affect (defaults to `downstream`)
+ - `toxicity`: probability of the toxic being applied to a link (defaults to 1.0, 100%)
+
+See [Toxics](#toxics) for available types. Each toxic type has its own parameters that
+can also be specified.
+
+The `stream` direction must be either `upstream` or `downstream`.
```
Sirupsen: Explain what this means and give an example of where it's significant.

README.md
xthexder: Done

CREATING_TOXICS.md
xthexder: When you `close()` a channel, all further reads will return nil. The toxic should never read more than one `nil` chunk, so I'll just change the wording here.

CREATING_TOXICS.md
xthexder: There really isn't anything the toxic needs to do other than make sure it doesn't drop any data. Th interrupt isn't really something a toxic can action upon, it's just the only way of killing the goroutine. I'm not really sure how to word this in the docs...

CREATING_TOXICS.md
xthexder: Added this above when `StreamChunk` is first referenced.

README.md
xthexder: Right now this is the only way to write custom toxics without merging upstream. It's an alternative to forking the whole project, which is what would have had to be done previously. I'll move this section into the `CREATING_TOXICS.md` file and just link to it, rather than having this weird summary.

Sirupsen:
Let me know when you've committed all the changes and I'll take another read

CREATING_TOXICS.md
xthexder: The Creating toxics section from the README was added here now.

xthexder:
I think that's all the chances except for the "what to do when an interrupt is received" section, which I'm still not entirely sure about.

Sirupsen:
Let's go with this

Sirupsen:
Bump :)

xthexder:
I can't really remember the status of this PR. Does "Let's go with this" mean merge?
As far as I can remember the docs for this are good to go.

Sirupsen:
Yep :+1: LGTM

PR #75, opened 2015-08-21
Add toxicity field

xthexder:
@Sirupsen @tomblench

Implements the `toxicity` field from https://github.com/Shopify/toxiproxy/pull/65 globally using the 2.0 toxic architecture.

Usage docs will appear in https://github.com/Shopify/toxiproxy/pull/74


Sirupsen:
Needs docs, otherwise LGTM

PR #76, opened 2015-08-24
2.0 Client Refactor

xthexder:
As I was documenting the go api, I had a look at a few projects that were using the go api, and almost all of them were wrapping the go client to make it easier to use. I took this as a sign that it needed a refactor.

Most operations are now a single line, rather than 2-3.

@Sirupsen @eapache


xthexder:
The toxiproxy-go client also needs a `Populate()` function like the ruby library has.

client/client.go
```diff
@@ -55,47 +62,37 @@ func (client *Client) Proxies() (map[string]*Proxy, error) {
...
-	return proxy
 }
 
-// Create creates a new proxy.
-func (proxy *Proxy) Create() error {
-	request, err := json.Marshal(proxy)
-	if err != nil {
-		return err
+// CreateProxy instantiates a new proxy and starts listening on the specified address.
+// This is an alias for `NewProxy()` + `proxy.Save()`
```
eapache: is there actually a need to expose `NewProxy` then?

eapache:
minor nit otherwise LGTM

client/client.go
xthexder: I left it exposed mostly because api_test.go uses it to create an already-disabled proxy, which is possible not really useful.
I'm not sure what you think about that use case, but I'd rather make it possible, rather than limiting things.

eapache: OK

client/example_test.go
```diff
@@ -0,0 +1,38 @@
...
+	// Test that redis is down
+}
+
+func TestRedisBackendSlow(t *testing.T) {
+	redisProxy.AddToxic("", "latency", "", toxiproxy.Toxic{
+		"latency": 1000,
+	})
+	defer redisProxy.RemoveToxic("latency")
+
+	// Test that redis is slow
```
Sirupsen: Is this a TODO?

Sirupsen:
Great call on checking out existing clients. Nothing really jumps out, it sounds like they all have a `CreateProxy` wrapper.

client/example_test.go
xthexder: This is a "Testing with toxiproxy" example, but it didn't realize it would be run with the toxiproxy tests until after.
I'll probably move this into the readme instead.

xthexder:
I've added the `Populate()` function, which works the same way as in the ruby library. I think these docs are fairly complete now.

Sirupsen:
This is great :)

PR #77, opened 2015-08-25
Make ChanReader interruptable

xthexder:
In order for `ChanReader` and `ChanWriter` to be used inside toxics, they need be interruptable.
This PR allows them to take the toxic's `interrupt` channel as a parameter, and unblock and reads.

@Sirupsen @eapache

### DIFF
--- a/stream/io_chan.go
+++ b/stream/io_chan.go
@@ -1,6 +1,7 @@
 package stream
 
 import (
+	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


xthexder:
See https://github.com/xthexder/toxictest/blob/master/http.go for how this would be used in a toxic.

eapache:
LGTM

PR #78, opened 2015-08-26
Clean up Makefile

xthexder:
Instead of using a hacky sed script to write the `VERSION` file into `version.go`, this uses the `-X` parameter of the linker.
I also made the default operation of `make` to be a dev build instead of a full release.

Dev builds will now show up as version `git-<sha>` instead of the previous release. If just `go build` is used, the version will be just `git`

@Sirupsen

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:1.4
 
 ADD . /app/src/github.com/Shopify/toxiproxy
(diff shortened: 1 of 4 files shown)


Sirupsen:
Awesome!

Sirupsen:
You can remove the "change toxiproxy.go" in the README as part of doing a release.

Sirupsen:
Could we make the Makefile also rewrite the README links so we can remove that step as well?

xthexder:
Without making a `README.md.erb` file or something, rewriting the versions in README would be pretty complicated. I don't think it's worth the time we'd save.

PR #80, opened 2015-08-28
Fix toxic marhsaling and toxicity updating

xthexder:
This code is kind of a mess, but it turns out there's no way to marshal a `ToxicWrapper` into a flat object without doing this hacky reflection.
The json package will marshal it as `{"toxic":{"latency": 1000}, "name": "latency")`, which isn't what we want.

Somehow this escaped the tests before, so I've added a bunch of checks to make sure it's covered.
While I was adding checks, I also found out that toxicitiy wasn't set when updating an existing toxic.

@Sirupsen @eapache

### DIFF
--- a/toxics/toxic.go
+++ b/toxics/toxic.go
@@ -1,6 +1,7 @@
 package toxics
 
 import (
+	"encoding/json"
 	"math/rand"
 	"reflect"
 	"sync"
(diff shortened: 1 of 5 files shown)


Sirupsen:
What is it we want instead? Can you show an example?

xthexder:
All endpoints that return a toxic should be marshalled like so:
`{"name":"latency","type":"latency","stream":"downstream","toxicity":1,"jitter":0,"latency":1000}`
i.e. in a flat map containing both the global and per-toxic fields.

Without this PR, the 2.0-dev branch is marshalling either just the global fields, or just the per-toxic fields depending on the endpoint, which is very broken.

toxics/toxic.go
```diff
@@ -32,14 +32,28 @@ type BufferedToxic interface {
...
+	Type       string
+	Stream     string
+	Toxicity   float32
+	Direction  stream.Direction
+	Index      int
+	BufferSize int
+}
+
+// Returns a flattened map of the toxic for use with json
+func (w *ToxicWrapper) GetMap() map[string]interface{} {
```
Sirupsen: Is this really better than defining a `JSONToxic` or similar and assigning to those fields? Not crazy about the usage of `reflect`.

xthexder: The problem is these fields are different for every toxic type. And since it's an interface, the json marshaller can't seem to figure out how to do nested structs. I'm really not happy with the usage of `reflect` either, so if you've got some other Go tricks up your sleeve to show me, I'd love to see them.

xthexder: It looks like we can improve this slightly by implementing `json.Marshaller` but that doesn't  get rid of the `reflect` package.

The problem `reflect` is solving is effectively this: http://play.golang.org/p/rZ0HQHhxSb

Sirupsen: Yeah, that sounds reasonable.

Sirupsen:
:+1:

toxics/toxic.go
pushrax: Note that internally, the JSON serializer is just using reflect too.

toxics/toxic.go
```diff
@@ -32,14 +32,28 @@ type BufferedToxic interface {
...
+	result := make(map[string]interface{})
+	ref := reflect.ValueOf(w.Toxic).Elem()
+	for i := 0; i < ref.NumField(); i++ {
+		result[ref.Type().Field(i).Tag.Get("json")] = ref.Field(i).Interface()
```
pushrax: This will assign one field with the name `-` afaik. See https://github.com/pushrax/flatjson/blob/master/flatjson.go#L32 for possible parsing logic.

pushrax:
What about changing the API?

toxics/toxic.go
xthexder: There was no `-` field when I tested this earlier and I'm not sure why there would be. You're right that this wouldn't handle `omitempty` though.

xthexder:
by changing the API I guess you mean just including a `toxic` sub-field with all the toxic-specific fields in it?

It could be done. It makes using curl a little more annoying, but if we get the cli released, that's not an issue.

toxics/toxic.go
pushrax: I mean, if `json:"-"` is used to specify that a field should be ignored in JSON output (http://play.golang.org/p/tPLcLLHZjE).

xthexder:
This now implements `json.Marshaller` instead of making our own `GetMap()` function.
Instead of using the reflect package, I've opted to just marshal twice and merge the output. That way we're sure to comply with all the `json:"-,omitempty"` weirdness.

Sirupsen:
LGTM

PR #81, opened 2015-10-19
Added skip feature for latency. 

mesutayata:
This will let skipping the effect of latency to the specified value by skip. Currently, the latency is being applied to all messages coming to the proxy. However, sometimes we may want to skip some messages not to be effected by the latency and some messages to be effected. This skip property is introduced to let any user to skip a certain amount of messages without any latency. It is  a percentage and should be between 0 and 100. If it is not set, by default it is 0, which means as usual all the messages coming to the latency proxy will be effected by latency. If it is 40, then roughly 40% of the messages will be skipped and 60% of messages will be applied this latency.

### DIFF
--- a/toxic_timeout.go
+++ b/toxic_timeout.go
@@ -1,6 +1,9 @@
 package main
 
-import "time"
+import (
+	"time"
+	"math/rand"
+)
 
 // The TimeoutToxic stops any data from flowing through, and will close the connection after a timeout.
 // If the timeout is set to 0, then the connection will not be closed.
@@ -8,6 +11,9 @@ type TimeoutToxic struct {
 	Enabled bool `json:"enabled"`
 	// Times in milliseconds
 	Timeout int64 `json:"timeout"`
+	// Percent of requests that will be passed without any effect of this timeout feature.
+	// Between 0-100 . 
+	Skip int32 `json:"skip"`
 }
 
 func (t *TimeoutToxic) Name() string {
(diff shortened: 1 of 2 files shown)


xthexder:
Unfortunately I would rather have this as part of the Toxiproxy 2.0 release (which has been a bit delayed recently).
In 2.0 a "toxicity" field was added that lets you control what percentage of connections are affected, which is somewhat similar to this.

toxic_timeout.go
```diff
@@ -23,7 +29,22 @@ func (t *TimeoutToxic) SetEnabled(enabled bool) {
...
+
+	// Percent of messages that will be passed without any effect of this timeout feature.
+	// When not set, skip = 0 by default, all messages are affected as usual.
+	skip := int32(t.Skip)	
+		
+	// Do not apply this timeout to this percent of packages.
+	if(skip>0){
+		var randValue = rand.Int31n(100);
+		if((randValue+1) <= skip){
+			return
```
xthexder: A toxic should only really return if it has received an interrupt or it is closing the connection. This will just cause the link to freeze.
You will need to pipe data through yourself like in the noop toxic if that's the intention here.

I also think in this case the `toxicity` field in 2.0 is exactly what you're looking to do here.

Sirupsen:
Agree, let's close this and get 2.0 shipped.

mesutayata:
That is awesome, Thank you for your interest and explanations :)

christian-esken:
Nice. This is one of the features I really missed in the existing toxiproxy version.

Sirupsen:
It's very close to being ready to go!

https://github.com/Shopify/toxiproxy/compare/master...2.0-dev

PR #83, opened 2015-11-02
Toxiproxy 2.0 Release

xthexder:
As far as I know, everything is ready to be released for 2.0

The released binary should be changed to `toxiproxy-server` or similar to allow for a cli binary later.
I've done some testing myself, but it would be good to get some verification that this works with the Shopify test suite.

Initial RFC is here: https://github.com/Shopify/toxiproxy/issues/54

PR's contained in this branch:
 * Initial PR https://github.com/Shopify/toxiproxy/pull/62
 * Remove stream direction from api urls https://github.com/Shopify/toxiproxy/pull/73
 * Update Readme https://github.com/Shopify/toxiproxy/pull/74
 * Add toxicity field https://github.com/Shopify/toxiproxy/pull/75
 * Go client refactor https://github.com/Shopify/toxiproxy/pull/76
 * Interruptable ChanReader https://github.com/Shopify/toxiproxy/pull/77
 * Makefile cleanup https://github.com/Shopify/toxiproxy/pull/78
 * Fix toxic marshalling and updating https://github.com/Shopify/toxiproxy/pull/80

These are all listed in the CHANGELOG.md

@Sirupsen @eapache @pushrax 

This shouldn't really need much review other than an overall :+1: and :tophat: 
Each of the above PRs have already been individually reviewed.

After this is merged, the `Release` steps just need to be followed (contained in the README)


Sirupsen:
@xthexder I'll get this working with Shopify today and report back. Let's run it in Shopify CI and developer environment for a week or two before a public release. How does that sound? We'll also have to update the Ruby client, I'll take care of that too.

Sirupsen:
@christian-esken @mesutayata @tomblench will all be interested in this release due to the toxicity field. Are you interested in testing this RC as well?

Sirupsen:
I've created a branch for `toxiproxy-ruby` to upgrade it to `2.0.0rc1`. First of all, this is so much better than Toxiproxy 1.0. Incredibly well done Jacob! 

Doing this independent integration testing a couple of questions came up:

1. `direction` was renamed to `stream`. Why do we think that's a better name?
2. The only reason for having a distinction between `name` and `type` I could come up with is if you ever wanted to apply multiple toxics of the same type. This can especially be useful for custom protocol-aware toxics, e.g. a 404-configured toxic on top of one that serves 5xxs. However, in my tests with the Latency toxic this stacking doesn't actually work. This [test fails](https://github.com/Shopify/toxiproxy-ruby/blob/2.0/test/toxiproxy_test.rb#L384-400) when it tries to create two Latency toxics and expects a delay that's the sum of the two.
3. Do we really need `GET /proxies/{proxy}/toxics` when we have `GET /proxies/{proxy}` also list the toxics?
4. Would it be simpler to return an array of toxics instead of a map with the names as keys?
5. Should we embed the `attributes` of each toxic in another attribute instead of                                                                                                                                    
  having a single namespace? This would prevent attributes for a toxic to leak into the "global" domain, which IMO is simpler to understand. E.g.                                                                                                                                                                                   
  ```go                                                                                                                                                                                                             
  ToxiFields struct {                                                                                                                                                                                               
    name string                                                                                                                                                                                                     
    type string                                                                                                                                                                                                     
    stream string                                                                                                                                                                                                   
    toxicity string                                                                                                                                                                                                 
    attributes map[string]string                                                                                                                                                                                    
  }                                                                                                                                                                                                                 
  ```
6. Should we change `GET /reset` to `POST /reset`? This works better with                                                                                                                                          standard HTTP where `POST` implies the mutation of content.
7. Why have `enable` / `disable` on proxies but not Toxics? (I think this makes                                                                                                                                       sense, just want to see your argument).
8. When submitting a toxic with a stream that doesn't exist you get a 404 saying: "stream not found". We should probably return a `HTTP 400` instead?

mesutayata:
@Sirupsen We will test some features soon, thanks for the notification !

xthexder:
Answers to questions:

1. Internally both `direction` and `stream` were used, so I basically just picked one so it's consistent. I don't really have much opinion either way.
2. Adding multiple toxics of the same type is definitely possible, but it looks like you found a bit of a bug with how latency delay is calculated... I'll see if I can get that fixed.
3. While `/toxics` isn't really necessary, it's a bit more REST-y. I don't see a problem with having both since it doesn't really add complexity.
4. That would make it a little simpler internally, I really don't know how the list of toxics is used externally though. That's something we can change if we want.
5. I guess the only reason I liked the flat structure was to make `curl` a bit cleaner to use, but if release the cli tool that doesn't matter anymore. It would certainly clean up the internal marshaling.
6. There was a bit of discussion about this before, but I think at this point `POST` makes more sense, so we can change that.
7. The reason we've got `enable/disable` for proxies is so you can turn off a proxy without losing all the existing toxic / port configuration, which you don't care about as much for a toxic. Basically it makes a `proxy.down()` block easier.
8. Yeah, that definitely makes more sense, we can change that.

Sirupsen:
1. :+1: 
2. I think there are two problems: 1) Latency toxic has a bug (as you pointed out), and 2) I don't think we actually want the `name` to default to the `type`, I think we want it to default to `{type}_{stream}`. Thoughts?
3. :+1: 
4. [ ] Let's look at this.
5. [ ] Yeah, I think we should optimize for implementation simplicity. It makes all other clients than `curl` awkward. Let's do this.
6. [x] https://github.com/Shopify/toxiproxy/issues/86
7. Agree, just wanted to make sure that was the only reason. :+1: 
8. [x] https://github.com/Shopify/toxiproxy/issues/85

Sirupsen:
CI passes for Shopify. I've just merged a PR that ships it to all our developers' environment. Let's see how it performs :-)

jpittis:
So it looks like I'm going to try to :ship: 2.0 :tada:!
First things first...
##### latency bug:
I think this should be default behaviour, though it needs documentation.

What's the "bug"? A latency proxy checks to see if a "packet" has been in the system longer than `latency` and delays it if so. If you have two latency toxics with the same `latency` value, the first one will hold the packet and the second will see that the packet has already been in the system for long enough and let it go.

##### Solutions:
* Toxic changes the arrival time after holding a packet. That sounds kinda sketchy and unexpected.

* We add some other packet state value to update by the latency toxic. Why should the latency toxic get a special piece of state like this?

* Keep some kind of latency specific state that keeps track of total delay time and use this value to increase latency toxics further down the pipe.

**Proposed solution:** Document it better. Any functionality that people would need 2 latency toxics for can be solved with one latency toxic, or a different type of toxic.

#### Todo's until :shipit: 
- [ ] Make a call on the latency "bug"
- [ ] Change default toxic name to `<type>_<stream>`
- [ ] Make frontend return toxics as a list see #92 
- [ ] Embed toxics into an attribute map
- [ ] `toxiproxy-client` :tada: #93 
- [ ] ruby client: @Sirupsen will do this close to :ship: time
- [ ] CHANGELOG also close to :ship: time

xthexder:
It makes sense to me that latency toxics should add up if you add multiple.
The way to fix this is to reset the timestamp on "packets" before sending them out of the latency toxic. As far as I know this should Just Work™
You can check the bandwidth toxic, since it does some stuff with the arrive time as well.
Arrival time is more of a time reference point to make things more accurate.

Sirupsen:
> It makes sense to me that latency toxics should add up if you add multiple.

Agree. The ability to do this to me is a sign of a healthy codebase.

xthexder:
@jpittis @Sirupsen I've actually got some time to work on toxiproxy this week.
I know there's a branch for the attribute map now, so that's possibly something I could take over?
`toxiproxy-client` probably shouldn't be updated until we have a finalized api.

Depending on how things go, I could also look at updating the ruby client, though I'm not sure if you have any specific plans for what that will look like @Sirupsen 

Let me know if I'm about to stomp all over your changes or if I'm good to start :ship:ing

Sirupsen:
@xthexder that sounds great to me! There's this branch https://github.com/Shopify/toxiproxy-ruby/pull/16, but it'll need to be upgraded along with this.

jpittis:
I'll PR the attribute map branch of tonight. I'd like to discuss a few of the changes with you. Once that's done, I believe that only the clients are left.

jpittis:
Decided not to PR because I thought of a better solution. Will post soon.

Sirupsen:
So basically we're done here.

@xthexder @jpittis do we want to start talking about 2.1 work, since it sounds like Jacob has some spare time? I think writing some sample protocol-aware toxics could be amazing. E.g. an HTTP toxic that can send 500s.

xthexder:
I could maybe look at that, though it's possible just getting 2.0 fully released will take up my time.
I'll be offline pretty much the whole weekend, but I can take a stab at an HTTP toxic next week for sure.

Sirupsen:
> though it's possible just getting 2.0 fully released will take up my time.

Yeah, let's focus on that first.

Once the two pending PRs are merged, we're good to go here. We'll need to test these in some production environments before releasing. @xthexder are you guys using it anywhere? I'll deploy it for Shopify's CI for a day or two. That will do millions of Toxiproxy calls.

Sirupsen:
We'll also ping the maintainers of other projects once the attribute branch gets merged in here, and we've confirmed things work as expected from updating the Ruby client.

jpittis:
@xthexder and @Sirupsen: The cli is merged! :tada:

We still need to clean the makefile and make sure the changelog and docs are in order.

Sirupsen:
This is SUPER exciting guys! So what we have left from what I can tell:

* [ ] Makefile updates
* [ ] Changelog and README
* [ ] Update Ruby client
* [ ] Perform actual release

xthexder:
I'll get to work on the Makefile and Changelog / README tomorrow, and hopefully the ruby client by the end of the week.
Definitely exciting

xthexder:
After this PR we just need the ruby-client update: https://github.com/Shopify/toxiproxy/pull/101 :100: :+1:

Sirupsen:
Both of you have been added as collaborators on `toxiproxy-ruby`. https://github.com/Shopify/toxiproxy-ruby/pull/16 just needs to be upgraded. I'm on vacation for the next ~2 weeks, so I can't test this in Shopify until I'm back. :/ But maybe you can get someone else to deploy it internally? Otherwise we can just release it when I'm back and it's been tested with Shopify.

xthexder:
2.0rc2 binaries uploaded here: https://github.com/Shopify/toxiproxy/releases/tag/v2.0.0rc2
Everything looks good to me, after `toxiproxy-ruby` is updated we should be good to go

Sirupsen:
@xthexder @jpittis I've put all of Shopify's internal development environments and CI on this change. I'm going to let it run for a week, if no issues, we'll move ahead with the release!

@mdevilliers (.NET maintainer), @ihsw (PHP maintainer), @dlion (Node.js maintainer), @trekawek (Java maintainer) please update your libraries! You can see the changes in [the changelog](https://github.com/Shopify/toxiproxy/blob/2.0-dev/CHANGELOG.md).

trekawek:
@Sirupsen, thanks for the notification. I create a [branch](https://github.com/trekawek/toxiproxy-java/tree/2.0-api) in the Java API client aligned to the 2.0 changes.

Sirupsen:
@trekawek great!

@jpittis @xthexder everything seems fine so far. I've set a reminder for Monday to release 2.0 if Shopify keeps running fine.

Sirupsen:
@xthexder @jpittis excellent work. I've released 2.0!

xthexder:
This is amazing! 2.0 finally done :100: 

I'll definitely have to have a play around with some custom http toxics and things now.

Sirupsen:
✨

jpittis:
:D

trekawek:
(y) Good job!

The [Java API client 2.0](https://github.com/trekawek/toxiproxy-java) with support for the new API format is available as well :)

PR #84, opened 2015-11-03
makefile: update for 2.0.0rc1

Sirupsen:
Updates the `Makefile` to compile `toxiproxy-server` and bump version to 2.0.0rc1. I used this for the internal Shopify build.

@xthexder


xthexder:
The Dockerfile needs an update to `toxiproxy-server` as well, but :+1:

Sirupsen:
I didn't think that really was worth it, do you disagree? E.g. Redis isn't called redis-server either on Docker Hub.

PR #85, opened 2015-11-03
api: improve stream error message

Sirupsen:
Per discussion in https://github.com/Shopify/toxiproxy/pull/83 

@xthexder

### DIFF
--- a/api.go
+++ b/api.go
@@ -347,7 +347,7 @@ var (
 	ErrMissingField       = newError("missing required field", http.StatusBadRequest)
 	ErrProxyNotFound      = newError("proxy not found", http.StatusNotFound)
 	ErrProxyAlreadyExists = newError("proxy already exists", http.StatusConflict)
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -347,7 +347,7 @@ var (
 	ErrMissingField       = newError("missing required field", http.StatusBadRequest)
 	ErrProxyNotFound      = newError("proxy not found", http.StatusNotFound)
 	ErrProxyAlreadyExists = newError("proxy already exists", http.StatusConflict)
-	ErrInvalidStream      = newError("stream not found", http.StatusNotFound)
+	ErrInvalidStream      = newError("stream was invalid, can be either upstream or downstream", http.StatusNotFound)
```
xthexder: Should we be changing this to http.StatusBadRequest?

xthexder:
:+1:

api.go
Sirupsen: Yes, forgot that!

PR #86, opened 2015-11-03
api: change reset method to be POST

Sirupsen:
As per discussion in https://github.com/Shopify/toxiproxy/pull/83

@xthexder

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -147,7 +147,7 @@ func (proxy *Proxy) Save() error {
 
 	var resp *http.Response
 	if proxy.created {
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -309,7 +309,7 @@ func (proxy *Proxy) RemoveToxic(name string) error {
 
 // ResetState resets the state of all proxies and toxics in Toxiproxy.
 func (client *Client) ResetState() error {
-	resp, err := http.Get(client.endpoint + "/reset")
+	resp, err := http.Post(client.endpoint+"/reset", "application/json", bytes.NewReader([]byte{}))
```
xthexder: Minor nit-pick, but "" isn't valid json. Maybe use `text/plain` ?

Issue #88, opened 2015-11-12
Deleting a timeout toxic can lockup proxy http requests


sunblaze:
Ran into this issue in some of my tests. Here's the curl commands you can use to recreate this issue.

```
curl -i -d '{"name": "redis", "upstream": "localhost:6379", "listen": "localhost:26379"}' localhost:8474/proxies
curl -i -d '{"type":"timeout", "timeout":30000, "stream": "upstream"}' localhost:8474/proxies/redis/toxics
curl -m 1 -i 127.0.0.1:26379
curl -m 1 -i -X DELETE localhost:8474/proxies/redis/toxics/timeout
```

The final delete request seems to lock up (final command must be executed before the 30 second timeout on the toxic), and any future requests like the following no longer returns before timing out

```
curl -i localhost:8474/proxies
```

But the version request works fine still `curl -i localhost:8474/version` returns 2.0.0rc1

@Sirupsen @xthexder

Sirupsen:
Oh, interesting. Thanks for the detailed bug report @sunblaze

xthexder:
I took a stab at this and found out what was going on. I wrote a fix, but in the process of testing found another bug with creating a new timeout toxic. There's some strange edge case with redis closing the connection and a toxic being added at the same time.

I should have a PR ready to fix this soon.

Sirupsen:
Fixed in 2.0

PR #89, opened 2015-11-15
Fix removing timeout toxic hang

xthexder:
Fixes https://github.com/Shopify/toxiproxy/issues/88

The order toxics were interrupted caused the api to hang, since the first toxic would try to flush data to the second (which in the case of a timeout toxic will never happen).
This PR changes the operation so that the timeout is interrupted first, and then make sure the previous toxic is able to complete by manually flushing data.

@Sirupsen @sunblaze @pushrax

### DIFF
--- a/link.go
+++ b/link.go
@@ -87,16 +87,21 @@ func (link *ToxicLink) Start(name string, source io.Reader, dest io.WriteCloser)
 
 // Add a toxic to the end of the chain.
 func (link *ToxicLink) AddToxic(toxic *toxics.ToxicWrapper) {
-	i := toxic.Index
+	i := len(link.stubs)
+
+	newin := make(chan *stream.StreamChunk, toxic.BufferSize)
+	link.stubs = append(link.stubs, toxics.NewToxicStub(newin, link.stubs[i-1].Output))
 
 	// Interrupt the last toxic so that we don't have a race when moving channels
 	if link.stubs[i-1].InterruptToxic() {
-		newin := make(chan *stream.StreamChunk, toxic.BufferSize)
-		link.stubs = append(link.stubs, toxics.NewToxicStub(newin, link.stubs[i-1].Output))
 		link.stubs[i-1].Output = newin
 
 		go link.stubs[i].Run(toxic)
 		go link.stubs[i-1].Run(link.toxics.chain[link.direction][i-1])
+	} else {
+		// This link is already closed, make sure the new toxic matches
+		link.stubs[i].Output = newin // The real output is already closed, close this instead
+		link.stubs[i].Close()
 	}
 }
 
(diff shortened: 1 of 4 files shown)


xthexder:
This PR still needs a test for the original issue, I'll add one a little later.

link.go
```diff
@@ -111,11 +116,40 @@ func (link *ToxicLink) UpdateToxic(toxic *toxics.ToxicWrapper) {
...
+						<-stop
+					}
+					return
+				}
+				link.stubs[i].Output <- tmp
+			}
+		}
+
 		// Empty the toxic's buffer if necessary
 		for len(link.stubs[i].Input) > 0 {
```
Sirupsen: Hm, do we really need this clause as well if we empty the channel above?

link.go
```diff
@@ -111,11 +116,40 @@ func (link *ToxicLink) UpdateToxic(toxic *toxics.ToxicWrapper) {
...
+		// Unblock the previous toxic if it is trying to flush
+		// If the previous toxic is closed, continue flusing until we reach the end.
+		interrupted := false
+		stopped := false
+		for !interrupted {
+			select {
+			case interrupted = <-stop:
+				stopped = true
+			case tmp := <-link.stubs[i].Input:
+				if tmp == nil {
```
Sirupsen: Do you really need the case above if you have this case? Or does it only return the end of the stream as `nil` once?

link.go
```diff
@@ -111,11 +116,40 @@ func (link *ToxicLink) UpdateToxic(toxic *toxics.ToxicWrapper) {
 func (link *ToxicLink) RemoveToxic(toxic *toxics.ToxicWrapper) {
 	i := toxic.Index
 
-	// Interrupt the last toxic so that the target's buffer is empty
-	if link.stubs[i-1].InterruptToxic() && link.stubs[i].InterruptToxic() {
+	if link.stubs[i].InterruptToxic() {
+		stop := make(chan bool)
+		// Interrupt the previous toxic to update its output
+		go func() {
+			stop <- link.stubs[i-1].InterruptToxic()
```
Sirupsen: What if we are interrupting the toxic with index 0?

Sirupsen:
@xthexder sorry about the delay, been away

link.go
xthexder: Index 0 is always a noop and cannot be removed.

link.go
xthexder: This part handles the case where the toxic has a buffer size != 0. The above case is only if the end of stream is reached while interrupting.
There might be a way to consolidate this into 1, I'll have a look later.

Sirupsen:
LGTM, simplification of the code for the part discussed would be nice, but not a must to be merged :+1:

link.go
xthexder: I'm not sure I know what you mean by this. This handles the case where the stream is closed before the toxic can be interrupted. We can also only call .Close() once or go will panic for closing an already closed channel.

Sirupsen: Ah ok, I see.

jpittis:
@xthexder are you alright with merging this into 2.0-dev?

PR #91, opened 2016-01-15
Cleanup cli and add 3 new commands

jpittis:
I thought I'd work on this.
**changes:**
1. Hard coded color constants were replaced with variables.
2. A `defer fmt.Printf(noColor)` was added to the beginning of `main` to return to default terminal colors on program exit.
3. Inlined command functions were [factored out](https://github.com/jpittis/toxiproxy/blob/46bb9443ac9fbbd8953c9a37986486a3cd9572c1/cli/cli.go#L32-L49) into the toplevel.
4. Create proxy, delete proxy and add toxic commands were added.

`create, c, new	Create a new proxy`
````
$ ./cli create myProxy 127.0.0.1:4444 127.0.0.1:7777
Created new proxy myProxy
````

`add, a		Add a toxic to a proxy`
````
$ ./cli add myProxy latency downstream '{"enabled":true,"latency":50,"jitter":30}'
Set downstream latency toxic on proxy myProxy
````

`delete, d		Delete a proxy`
````
$ ./cli delete myProxy
Deleted proxy myProxy
````

**I few things I havn't done yet that I think need doing:**
1. We need better command discriptions and examples.
2. Removing toxics wasn't implemented beccause the go client doesn't implement very good toxic manipulation. That can be easily fixed too.
@Sirupsen, @xthexder: Is this in the right direction?

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"encoding/json"
+	"log"
+	"sort"
+
 	"github.com/Shopify/toxiproxy/client"
 	"github.com/codegangsta/cli"
-	"sort"
 
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cli/cli.go [hunk 1]
```diff
@@ -29,86 +33,180 @@ func main() {
...
+			err.Error())
+	}
+	fmt.Printf("Deleted proxy %s\n", proxyName)
+}
+
+func addToxic(c *cli.Context, t *toxiproxy.Client) {
+	proxyName := c.Args().Get(0)
+	toxicName := c.Args().Get(1)
+	direction := c.Args().Get(2)
+	toxicConfig := c.Args().Get(3)
```
Sirupsen: I think we should use named arguments instead. How does this look in the help pages?

Sirupsen:
<3 for taking this over

cli/cli.go
jpittis: Named arguments (options) would look something like this:
````
$ ./cli create --help
NAME:
   cli create - Create a new proxy

USAGE:
   cli create [command options] [arguments...]

OPTIONS:
   --name, -n 		The name of the proxy.
   --listen, -l 	Proxy will listen on this address.
   --upstream, -u    Proxy will forward to this address.
```
```
./cli create -n myProxy -l 127.0.0.1:7777 -u 127.0.0.1:4444
```
I prefer using named arguments.

jpittis:
9370001 through 89fd627
* Uses named arguments of the style:
````
./cli set -n myProxy -t latency -f '{"latency":50,"enabled":true}' -d
````
* Inspect displays headings for upstream and downstream toxics:
````
...
latency direction=upstream enabled=true latency=100 jitter=0

downstream
bandwidth direction=downstream enabled=false rate=0
...
````
* Toxics can be `set` on either downstream with `-d`, upstream `-u` **or both!!**.

**TODO**
* Not specfying named arguments like `--name` will lead to http requests with an empty name field. Printed error may be unclear.
* Fancy char symbols that show what toxics are currently activated.

jpittis:
57456d8 checks for required named arguments
````
$ ./cli i
2016/01/20 10:52:02 Required argument 'name' was empty.
````

faa54de adds pipe drawings to inspect command
````
$ ./cli i -n myProxy
myProxy
127.0.0.1:4444 --L-SC-T-B--> 127.0.0.1:5555
````

@Sirupsen: I'm thinking that's about all this needs to ship. I tried to break the commits up so they are easy to follow. We might want to squash later.

Logging prints a date and time because I'm using `log.fatal`. We might prefer switching back to `fmt.Print + os.Exit`.

Sirupsen:
I need to play around with the UX here to vouch for it, I'll try to do that before Monday.

Sirupsen:
Notes from playing around with this:

* `list` should have a column that says up / down which should be coloured. The colour on the name isn't intuitive enough.
* `set` should probably be renamed to something like `toxic`
* `set` should list all the available toxics
* `set` should default to downstream
*  `set` syntax for setting parameters should probably be in `key=value` form, e.g. `toxiproxy-client set --name redis --toxic latency --fields "latency=20"`
* `set` should enable a toxic by default (`--enable` flag) and then allow you to disable it (`--disable`)
* `set` should default to `enabled=true` so you don't have to specify it to `fields`
* `set` should have a full example when no argument is supplied
* `list `should show if a proxy has toxics enabled
* `inspect` overview is pretty cool, but should probably be a little more descriptive, e.g. `name: <proxy_name>` and `Upstream Toxics` and `Downstream Toxics`.

This is awesome though Jake. So happy about this!

jpittis:
Closing this PR. Was rebased on to dev-2.0 here: #93

PR #92, opened 2016-01-27
api: toxics are sent to client as array rather than map

jpittis:
> Would it be simpler to return an array of toxics instead of a map with the names as keys?

This changes the api to return a list of toxics rather than a map of name to toxic.

I did not change how toxics are implemented internally. Instead, this is a front end change.

@Sirupsen and @xthexder were unsure about whether this is needed.

## Before
```
"toxics": {
    "myToxic1": {"name": "myToxic1"...},
    "myToxic2": {"name": "myToxic2"...}
}
```
## After
```
"toxics": [
    {"name": "myToxic1"...},
    {"name": "myToxic2"...}
]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Make `ChanReader` in the `stream` package interruptible #77
 * Define proxy buffer sizes per-toxic (Fixes #72)
 * Fix slicer toxic testing race condition #71
+* API and client return toxics as array rather than a map of name to toxic #92
 
 # 1.2.1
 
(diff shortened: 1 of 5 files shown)


Sirupsen:
Great! This'll require changes in the Ruby clients as well, otherwise looks good.

Could this be added to the CHANGELOG where we'll do a best effort collection of backwards incompatible changes? I'll contribute to it what I learned from implementing the Ruby client once we're closer to a release.

jpittis:
Would you confirm 6b94ceb has the right style.

Sirupsen:
LGTM

PR #93, opened 2016-02-01
Cli 2.0 dev

jpittis:
Here's `toxiproxy-client` v2.0!

- [x] I should close both #58 and #91 as this PR replaces them.

Not worrying about clear commit history when doing rapid ux changes.
I'll start with a quick list of changes, followed by relevant screen shots.
1. Updated version number to 2.0
2. Removed graphic pipe in `inspect` because I found it confusing
3. Made `inspect` a lot clearer
4. Made `list` a lot clearer. Also table has `Toxic` and `Enabled` column
5. Added cyan hints (might be kinda silly?)
6. Added usage and examples to toplevel command and `toxic` command
7. Updated the `makefile` (@Sirupsen: would you glance at this, I'm a little rusty with make)

### Screen Shots
Part of the toplevel help screen.
toplevel-help
Part of the `./cli toxic` screen.
toxic-help
Looks like we havn't created any toxics yet...
ls-empty
There's one!
ls-one
Let's try to inspect it!
inspect-one
No toxics? Let's add one!
inspect-down

Darn! It defaulted to downstream... let's add another upstream!
inspect-up-down
And let's list the proxies again. **(This is after I added the enabled field)**
with-enabled

TODO:
- [x] #92 merge will require an update
- [ ] do I need to update the CHANGELOG?
- [x] ux review from @Sirupsen 
- [x] a ux review from someone who knows toxiproxy but not the client
- [x] Redo inspect command
- [ ] Configure toxiproxy host ip.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"sort"
+	"strconv"
+	"strings"
+
 	"github.com/Shopify/toxiproxy/client"
 	"github.com/codegangsta/cli"
-	"sort"
 
 	"fmt"
 	"os"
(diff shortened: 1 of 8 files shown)


Sirupsen:
@es mind doing a UX review here as well since you're interested in Toxiproxy and don't know too much about it?

xthexder:
:100: this looks amazing.

One thing I would point out is that the field order on toxics looks to be random (go maps do that by default)
It would be nice if you sorted them or something to make it consistent.

Other than that :+1: 
I'd love to work on Toxiproxy some more, but unfortunately it's going to have to be limited to weekends since I'm busy working on other things.

es:
This looks intuitive except for the `-n` flag in `./cli inspect -n <PROXY_NAME>`. `./cli inspect <PROXY_NAME>` seems more intuitive to me 

:wave: UX :wave:

xthexder:
I agree with Emil, it seems more intuitive to skip the `-n` flag and just have the proxy name as the next argument.

Sirupsen:
Did another pass of nitpicks:

* Piping to a non-TTY (e.g. a file) shouldn't print the colors
* The white colour for the `====` separator is very bright, let's make it normal text colour
* It'd be more helpful if the default when typing `toxiproxy-client create` would be the `--help` screen, instead of "Required argument name is missing"
* Agree with @es and @xthexder on skipping the `--name` flag consistently for the proxy name, which is required to pretty much everything
* Is setting a toxic really the right terminology? Doesn't `add` make more sense?
* Print the hints in neutral colour instead of highlighting it. The colours are for helping guide vision, not to create a party :-)
* The `inspect` output could be more nicely structured. Exactly how I don't have a good idea, but I think it'd be worth it to play around with.

Lastly, we should also mark the client as being alpha, so we can break backwards compatibility as we iterate on its usage and get feedback—instead of having it under the same semantic versioning as the server, which is fairly stable—especially with 2.0.

Overall this is looking amazing Jake. Thanks for taking the prototype and running with it.

jpittis:
Made all the suggested changes except:
1. Sorting the toxics before printing them. (I believe #92 fixes this.)
- [ ] A way to set the toxiproxy-server port.

Here's my attempt to make inspect cleaner @Sirupsen:
screenshot from 2016-02-09 14-40-07

cli/cli.go
```diff
@@ -0,0 +1,466 @@
...
+
+func listToxics(toxics toxiproxy.Toxics, stream string) {
+	fmt.Printf("%s%s toxics:\n%s", greenColor, stream, noColor)
+	if len(toxics) == 0 {
+		fmt.Printf("%sProxy has no %s toxics enabled.\n%s", redColor, stream, noColor)
+		return
+	}
+	for name, toxic := range toxics {
+		fmt.Printf("%s%s:%s", greenColor, name, noColor)
+		for property, value := range toxic {
```
xthexder: I think the field order will still be random, even with the changes in #92 
It would be nice to have a fixed order these.

Sirupsen:
Then I think we're good here :+1:

Sirupsen:
Once @xthexder's comment is addressed here, I think this is ready to make it into 2.0.

jpittis:
Looks like I mucked up something with git. fix coming soon :tm:

jpittis:
Nevermind... ah I hate having to force push rebased branches.

@xthexder: 8329c1b makes the client use the new nested toxic API and sorts the attributes list. Also `AddToxic` and `UpdateToxic` do some footwork to make toxicity work well. A review would be appreciated. :)

jpittis:
Do you mean this sortof thing @xthexder? 00f5367

xthexder:
Not quite what I had in mind, though it does work. I was thinking we'd have `UpdateToxic()` in the go client handle it, and just not send the field to the server if it's -1.
Just had a look at that code and it looks like we're sending something like `{"name": "", "type": "", "toxicity": 1.0, "attributes": {"latency": 1000}}` as it is, wouldn't be a bad idea to just marshal a `map[string]interface{}` instead and only include the fields we want (attributes, and optionally toxicity)

xthexder:
I went ahead and pushed a commit for what I mentioned. The `Makefile` is going to need a bit of work before this is merged as well.
I personally would perfer `toxiproxy-cli` at this point, considering this PR and the code folder are named cli (at there is already an unrelated client folder).
Any comments on if we should just ship 1 .deb package with both cli + server?

client/client.go
```diff
@@ -240,6 +240,9 @@ func (proxy *Proxy) Toxics() (Toxics, error) {
 // See https://github.com/Shopify/toxiproxy#toxics for a list of all Toxic types.
 func (proxy *Proxy) AddToxic(name, typeName, stream string, toxicity float32, attrs Attributes) (*Toxic, error) {
 	toxic := Toxic{name, typeName, stream, toxicity, attrs}
+	if toxic.Toxicity == -1 {
```
Sirupsen: Why is this better than just passing 1.0 by default from the CLI?

xthexder: Well.. it isn't. I put it there just so -1 can be used as the default anywhere and to be consistent. Maybe that's not something we care about though.

jpittis:
The `-1` stuff LGTM

You could probably do a much better job than me cleaning the Makefile. Would you be able to do that? I guess we're good to ship after that!

xthexder:
Sure, I can do the Makefile changes, would maybe be better suited to another PR though.
We can merge this unless you think we should change the `-1` thing @Sirupsen

Sirupsen:
Doesn't matter too much to me. I'm good with this!

jpittis:
@xthexder  Should I squash all of my commits togeather? Whats the best practice?

Sirupsen:
yeah, squashing sounds great

PR #94, opened 2016-02-02
latency toxic should reset Timestamp on output

jpittis:
Multiple latency toxics previously did not stack their delays. Latency toxics now reset their data's `Timestamp` before output.

**Pros:** Easy fix that *just works*
**Cons:** Changing `Timestamp` state that might want to be used by other toxics in the future.

*Decided to create a new test rather than make the original latency test support multiple toxics.*

Sorry to ping you again @xthexder but you seem to know the most about this.
Also @Sirupsen, see https://github.com/Shopify/toxiproxy/pull/83#issuecomment-178224768 and https://github.com/Shopify/toxiproxy/pull/83#issuecomment-178221647

### DIFF
--- a/toxics/latency.go
+++ b/toxics/latency.go
@@ -39,8 +39,10 @@ func (t *LatencyToxic) Pipe(stub *ToxicStub) {
 			sleep := t.delay() - time.Since(c.Timestamp)
 			select {
 			case <-time.After(sleep):
+				c.Timestamp = c.Timestamp.Add(sleep)
 				stub.Output <- c
 			case <-stub.Interrupt:
+				// Exit fast without applying latency.
 				stub.Output <- c // Don't drop any data on the floor
 				return
 			}
--- a/toxics/latency_test.go
+++ b/toxics/latency_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"io"
 	"net"
+	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


Sirupsen:
Makes sense to me :+1:

toxics/latency.go
```diff
@@ -37,10 +37,14 @@ func (t *LatencyToxic) Pipe(stub *ToxicStub) {
 				return
 			}
 			sleep := t.delay() - time.Since(c.Timestamp)
+			start := time.Now()
 			select {
 			case <-time.After(sleep):
+				c.Timestamp = c.Timestamp.Add(sleep)
 				stub.Output <- c
 			case <-stub.Interrupt:
+				slept := time.Now().Sub(start)
```
xthexder: I'm not a big fan of using a timed value here, though in this instance I think it does work.
When the toxic is interrupted, it's a bit of an edge-case where we don't really care too much about the latency, just that it gets through.
It's up to you if you want to leave this here or not.

xthexder:
:+1: Just the one thing to say. It's more of a gut feeling I have rather than any technical reason..

jpittis:
Now that you mention it @xthexder, when experiencing an interupt, we want the toxic to exit quickly. For example, latency needs not be applied while adding new toxics to a proxy.

PR #98, opened 2016-03-16
Embed toxic attribute

jpittis:
This patch embeds toxic specific attributes into their own nested `attributes` field, allowing for the following api changes:
``` go
type Attributes map[string]interface{}

type Toxic struct {
	Name       string     `json:"name"`
	Type       string     `json:"type"`
	Stream     string     `json:"stream"`
	Toxicity   float32    `json:"toxicity"`
	Attributes Attributes `json:"attributes"`
}

AddToxic(name, typeName, stream string, toxicity float32, attrs Attributes)
UpdateToxic(name string, toxicity float32, attrs Attributes)
```
We can't marshal the attributes of a toxic until we know what kind of toxic struct to marshal into. This has previously forced us to [parse the json twice](https://github.com/Shopify/toxiproxy/blob/d21a2744e6275eb5312b11ab978656fd7cad50bc/toxic_collection.go#L95).

I can't think of a better solution than to [use an anonymous struct](https://github.com/Shopify/toxiproxy/blob/d21a2744e6275eb5312b11ab978656fd7cad50bc/toxic_collection.go#L129-L138) to extract the specific json that we need.

This has also caused the previous toxic marsheling to be kinda sketchy. [My change does not make it less sketchy](https://github.com/Shopify/toxiproxy/blob/d21a2744e6275eb5312b11ab978656fd7cad50bc/toxics/toxic.go#L57-L60).

I'm happy enough with the approach. There might not be a pretty solution. @Sirupsen or @xthexder might be able to think of one?

Update *CHANGELOG*, docs and squish the commits once we're ready to merge.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -17,7 +17,16 @@ type Client struct {
 	endpoint string
 }
 
-type Toxic map[string]interface{}
+type Attributes map[string]interface{}
+
+type Toxic struct {
+	Name       string     `json:"name"`
+	Type       string     `json:"type"`
+	Stream     string     `json:"stream,omitempty"`
+	Toxicity   float32    `json:"toxicity"`
+	Attributes Attributes `json:"attributes"`
+}
+
 type Toxics []Toxic
 
 // Proxy represents a Proxy.
@@ -226,22 +235,13 @@ func (proxy *Proxy) Toxics() (Toxics, error) {
 }
 
 // AddToxic adds a toxic to the given stream direction.
-// If a name is not specified, it will default to the same as the type.
+// If a name is not specified, it will default to <type>_<stream>.
 // If a stream is not specified, it will default to downstream.
 // See https://github.com/Shopify/toxiproxy#toxics for a list of all Toxic types.
-func (proxy *Proxy) AddToxic(name, typeName, stream string, toxic Toxic) (Toxic, error) {
-	if toxic == nil {
-		toxic = make(Toxic)
-	}
-	toxic["type"] = typeName
-	if name != "" {
-		toxic["name"] = name
-	}
-	if stream != "" {
-		toxic["stream"] = stream
-	}
+func (proxy *Proxy) AddToxic(name, typeName, stream string, toxicity float32, attrs Attributes) (*Toxic, error) {
+	toxic := Toxic{name, typeName, stream, toxicity, attrs}
 
-	request, err := json.Marshal(toxic)
+	request, err := json.Marshal(&toxic)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


xthexder:
`MarshalJSON()` can be removed entirely with this change:

```
type ToxicWrapper struct {
	Toxic      `json:"attributes"`
	Name       string           `json:"name"`
	Type       string           `json:"type"`
	...
}
```

The custom `MarshalJSON` was specifically to flatten the toxic into a single map, which no longer needs to be done. The double json parse isn't ideal, but I'm not really seeing many other options to make it better.

client/client.go
```diff
@@ -282,13 +283,13 @@ func (proxy *Proxy) UpdateToxic(name string, toxic Toxic) (Toxic, error) {
 		return nil, err
 	}
 
-	toxics := make(Toxic)
-	err = json.NewDecoder(resp.Body).Decode(&toxics)
+	result := &Toxic{}
+	err = json.NewDecoder(resp.Body).Decode(&result)
```
xthexder: Should remove the `&` here, it's a double pointer otherwise.

client/client.go
```diff
@@ -17,7 +17,16 @@ type Client struct {
 	endpoint string
 }
 
-type Toxic map[string]interface{}
+type Attributes map[string]interface{}
+
+type Toxic struct {
+	Name       string     `json:"name"`
+	Type       string     `json:"type"`
+	Stream     string     `json:"stream"`
```
xthexder: We can make use of `json:"stream,omitempty"` here for default values.
I'd like to continue allowing toxics to be of toxicity 0, at least in the server, but I'm not sure what to do about that on the client. Maybe we just don't have a "default" value in the client.

jpittis:
> `MarshalJSON()` can be removed entirely with this change:

I don't believe it can be.

When adding the `json:"attributes"`, unmarshaling the json will fail because it tries to unmarshal the attributes map into a nil `Toxic struct`. We cannot set the nil `Toxic` [until we know the toxic's type](https://github.com/Shopify/toxiproxy/blob/d21a2744e6275eb5312b11ab978656fd7cad50bc/toxic_collection.go#L117-L119). We need to unmarshal the json to find the toxic's type.

xthexder:
I guess I didn't try it to see what would happen, but we could maybe just temporarily set the toxic as a NoopToxic and then redefine it later. This isn't the best way of doing things, but it's still much better than leaving the `MarshalJSON()` function around. This also means we don't really need the anonymous structs for the second unmarshal.

toxics/toxic_test.go
```diff
@@ -99,30 +99,22 @@ func ToxicToJson(t *testing.T, name, typeName, stream string, toxic toxics.Toxic
...
-		}
-		return nil
-	}
-	marshal["type"] = typeName
-	if name != "" {
-		marshal["name"] = name
-	}
-	if stream != "" {
-		marshal["stream"] = stream
+	data := map[string]string{
```
xthexder: This can just be `map[string]interface{}` with `"attributes": toxic,`

xthexder: Actually, this would be another option for `MarshalJSON()` if we do keep it around, would be much cleaner at least.

client/client.go
jpittis: I'll add a test to pick this up.

client/client.go
jpittis: To use omniempty we need to make `Stream` an `&string` right? Because a strings zero value is `""`?

xthexder: It's fine to leave it as a `string`. The idea being you pass in `""` to `AddToxic()` and then get the default stream of `"downstream"` and we don't send a blank string in the json.

Sirupsen:
> I guess I didn't try it to see what would happen, but we could maybe just temporarily set the toxic as a NoopToxic and then redefine it later. 

This sounds great if it works.

jpittis:
The Noop works perfectly!

Sirupsen:
This looks really clean, awesome!11!!

xthexder:
Looks great! Should be good to merge into 2.0-dev with a CHANGELOG entry.

jpittis:
Probably need some changes to the docs too.

xthexder:
I went ahead and squashed + rebased the branch, so it's up to date with `2.0-dev`
I've updated the docs as well now to include the `<type>_<stream>` and nested attribute changes

Should be good to merge after one more :+1: from everyone.

jpittis:
Wont be able to look at this until tomorrow afternoon.

Sirupsen:
A++++ ship!

Sirupsen:
@xthexder if you want to own updating the Ruby branch (https://github.com/Shopify/toxiproxy-ruby/pull/16), go for it.

jpittis:
:+1: :D

Issue #102, opened 2016-04-18
Batch API / support HTTP pipelines


sirupsen:
@BlakeMesdag mentioned that creating proxies at boot of our application actually takes a while, up to 2 seconds. We can make this better. Toxiproxy should not take more than 100ms at boot, and preferably less.

The simplest thing would probably be to support HTTP pipelines / persistent connections, so we can just send a bunch of data. We may also just be doing something stupid in the Rails gem. I'd prefer that to a new API. Another possibility is that the `#populate` call becomes a native thing in Toxiproxy that just sends it a JSON dictionary of all the proxies it expects (and resets all of them).

@xthexder @jpittis

xthexder:
A `/populate` endpoint wouldn't be a bad thing to implement. It would easily be able take advantage of concurrent proxy creation (which is currently only possible with parallel requests).
Maybe something for a 2.1 release?

I noticed @BlakeMesdag already started a branch, is there any performance improvement with that?

BlakeMesdag:
@xthexder the branch is just switching from manually creating new requests and passing them to the connection, to using the connection's instance methods for those requests.

Sirupsen:
\cc @jpittis this might be something you'd be interested in

jpittis:
FYI That I've started working on a populate fix.

Sirupsen:
@xthexder are you interested in taking this up? @jpittis did work in #111 to add a populate endpoint, but it seems the problem is not with performance of all the network roundtrips (although I do feel we should have `/populate` anyway) but a performance problem inside of Toxiproxy.

xthexder:
Yeah I can take a look at this and do some benchmarks to see what the problem is.

xthexder:
Should be fixed with the Toxiproxy 2.1.0 release.

PR #111, opened 2016-06-18
Add a populate endpoint

jpittis:
Here's a WIP first draft for a Populate endpoint solving #102.
(The code can be made prettier and some names / types need to be changed to stick with the old API.)

The endpoint takes a slice of proxies, loop through them doing the standard CreateProxy code. If one of the proxies fails to be created, it deletes the already created proxies before returning the error to the user.

**Should we be making this transactional? There are current errors that can occur that won't trigger the transaction to fail.**

Hoping I could get a review of the general approach.
@Sirupsen @xthexder 
cc: @BlakeMesdag

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ stopping you from creating a client in any other language (see
   1. [Proxy fields](#proxy-fields)
   2. [Toxic fields](#toxic-fields)
   3. [Endpoints](#endpoints)
+  4. [Populating Proxies](#populating-proxies)
 7. [CLI example](#cli-example)
 8. [FAQ](#frequently-asked-questions)
 9. [Development](#development)
@@ -440,6 +441,16 @@ All endpoints are JSON.
  - **DELETE /proxies/{proxy}/toxics/{toxic}** - Remove an active toxic
  - **POST /reset** - Enable all proxies and remove all active toxics
  - **GET /version** - Returns the server version number
+
+#### Populating Proxies
+
+Proxies can be added and configured in bulk using the `/populate` endpoint. This is done by
+passing an json array of proxies to toxiproxy. If a proxy with the same name already exists,
+it will be compared to the new proxy and replaced if the `upstream` and `listen` address don't match.
+
+A `/populate` call can be included for example at application start to ensure all required proxies
+exist. It is safe to make this call several times, since proxies will be untouched as long as their
+fields are consistent with the new data.
 
 ### CLI Example
 
--- a/README.md
+++ b/README.md
@@ -429,6 +429,7 @@ All endpoints are JSON.
 
  - **GET /proxies** - List existing proxies and their toxics
  - **POST /proxies** - Create a new proxy
+ - **POST /populate** - Create or replace a list of proxies
  - **GET /proxies/{proxy}** - Show the proxy with all its active toxics
  - **POST /proxies/{proxy}** - Update a proxy's fields
  - **DELETE /proxies/{proxy}** - Delete an existing proxy
--- a/api.go
+++ b/api.go
@@ -27,6 +27,7 @@ func (server *ApiServer) Listen(host string, port string) {
 	r.HandleFunc("/reset", server.ResetState).Methods("POST")
 	r.HandleFunc("/proxies", server.ProxyIndex).Methods("GET")
 	r.HandleFunc("/proxies", server.ProxyCreate).Methods("POST")
+	r.HandleFunc("/populate", server.Populate).Methods("POST")
 	r.HandleFunc("/proxies/{proxy}", server.ProxyShow).Methods("GET")
 	r.HandleFunc("/proxies/{proxy}", server.ProxyUpdate).Methods("POST")
 	r.HandleFunc("/proxies/{proxy}", server.ProxyDelete).Methods("DELETE")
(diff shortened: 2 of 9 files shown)


Sirupsen:
@jpittis what have you located as the source of the slowness? Is it in the sheer # of HTTP connections for creating each proxy (this would fix it), or is it Toxiproxy actually being slow (this wouldn't fix it)?

jpittis:
This is going to take a bit more work than I expected. Here are some basic benchmark stats showing that my naive fix was a little too naive.
```
$ go test -bench .
PASS
BenchmarkPopulate-8   	     100	 126641038 ns/op
BenchmarkPopulateNew-8	     100	 136556076 ns/op
ok  	github.com/Shopify/toxiproxy	26.706s
```

Sirupsen:
@jpittis yeah so it seems like it's legitimately taking a long time to create each proxy, and that the performance is not with the TCP/IP overhead? Good excuse to dig into some Go profile tooling for you. :)

xthexder:
Without making any changes to the branch, and benchmarking `.Populate()` vs `.Populate2()` the new api call is about 4x faster than individual calls.

I'm able to create 5000 proxies in ~0.8s. any more and I run out of ports. The profiler basically ends up with 90% syscalls, so there's not really anywhere we can go from here.
```
go test -run=none -bench=. -count=1 -benchtime=300ms -cpuprofile=cprof .
PASS
BenchmarkPopulateProxyList2-8  	    5000       	    127590 ns/op
ok     	github.com/Shopify/toxiproxy   	0.802s
```

The old one takes quite a bit longer and has a decent amount of time spent accepting TCP and writing to the toxiproxy client.
```
go test -run=none -bench=. -count=1 -benchtime=2s -cpuprofile=cprof .
PASS
BenchmarkPopulateProxyList-8   	    5000       	    400418 ns/op
ok     	github.com/Shopify/toxiproxy   	2.089s
```

I wouldn't mind taking a look at the benchmarking code @jpittis wrote, but I think this is probably good to merge after a bit of cleanup.

xthexder:
Also, here's the benchmark numbers for a few less proxies:
```
go test -run=none -bench=. -count=1 -benchtime=60ms -cpuprofile=cprof .
PASS
BenchmarkPopulateProxyList-8   	     200       	    340446 ns/op
BenchmarkPopulateProxyList2-8  	    1000       	     87665 ns/op
ok     	github.com/Shopify/toxiproxy   	0.235s
```

Sirupsen:
:metal:

xthexder:
Well, I pushed some changes that clean things up a bit and fix a couple issues like the default `enabled` value.

Main things changed:
 - `/populate` is no longer transactional. I don't think it's worth the cost of code complexity, and this way it operates the same as the existing clients.
 - `/populate` will return the first error encountered, along with the proxies successfully created before the error.
 - The `ApiError` object was changed from `{"title": "foo"}` to `{"error": "foo"}` to make things less confusing.
 - Added a whole bunch of tests.

Should be good for a review @Sirupsen @jpittis

proxy_collection.go
```diff
@@ -38,6 +38,29 @@ func (collection *ProxyCollection) Add(proxy *Proxy, start bool) error {
 	return nil
 }
 
+func (collection *ProxyCollection) AddOrReplace(proxy *Proxy, start bool) error {
+	collection.Lock()
+	defer collection.Unlock()
+
+	if existing, exists := collection.proxies[proxy.Name]; exists {
+		if existing.Listen == proxy.Listen && existing.Upstream == proxy.Upstream {
```
xthexder: There's a slight problem with this, since toxiproxy will rewrite `localhost` to `127.0.0.1` and these won't always match as expected. I may put out another PR later that does a proper resolve test/compare.

The logic here is that `/populate` should be safe to rerun several times without resetting all the proxies, which currently only works if the input is static IP addresses.

BlakeMesdag:
@xthexder would it be better to process the full list of proxies and return a list of errors for each one?

Having it stop half-way (or wherever it breaks) might be more annoying since you can't know which order toxiproxy parsed them. We could assume that it will always be the same as the JSON payload, but I don't know if that's guaranteed.

xthexder:
I originally started implementing it that way, but the format for returning multiple errors, especially in the client side gets messy. As it is, the order is guaranteed to be the same as the json array (would not be the case for a map)

One of the problems is determining which error is for which proxy on the client. The go client could return `([]Proxy, []error)` but for ruby I don't know what we'd have to do since the errors are raised as exceptions.

api.go
```diff
@@ -130,6 +131,74 @@ func (server *ApiServer) ProxyCreate(response http.ResponseWriter, request *http
...
+			apiErr, ok = err.(*ApiError)
+			if !ok {
+				logrus.Warn("Error did not include status code: ", err)
+				apiErr = &ApiError{err.Error(), http.StatusInternalServerError}
+			}
+			responseCode = apiErr.StatusCode
+			break
+		}
+
+		proxies = append(proxies, proxyWithToxics(proxy))
```
Sirupsen: What is this `proxyWithToxics` helper for? Why aren't they are on the proxy object already?

api.go
```diff
@@ -130,6 +131,74 @@ func (server *ApiServer) ProxyCreate(response http.ResponseWriter, request *http
...
+
+	responseCode := http.StatusCreated
+	var apiErr *ApiError
+	for _, p := range input {
+		proxy := NewProxy()
+		proxy.Name = p.Name
+		proxy.Listen = p.Listen
+		proxy.Upstream = p.Upstream
+
+		err = server.Collection.AddOrReplace(proxy, *p.Enabled)
```
Sirupsen: This method was there already? Was it used for the `POST` API, or why?

api.go
```diff
@@ -130,6 +131,74 @@ func (server *ApiServer) ProxyCreate(response http.ResponseWriter, request *http
...
+
+	// Check for valid input before creating any proxies
+	t := true
+	for i, p := range input {
+		if len(p.Name) < 1 {
+			apiError(response, joinError(fmt.Errorf("name"), ErrMissingField))
+			return
+		}
+		if len(p.Upstream) < 1 {
+			apiError(response, joinError(fmt.Errorf("upstream"), ErrMissingField))
```
Sirupsen: These errors don't say anything about the name of the proxy where this error occurs the way I read it. That may be unhelpful, what do you think?

api.go
```diff
@@ -130,6 +131,74 @@ func (server *ApiServer) ProxyCreate(response http.ResponseWriter, request *http
 	}
 }
 
+func (server *ApiServer) Populate(response http.ResponseWriter, request *http.Request) {
+	input := []struct {
+		Proxy
+		Enabled *bool `json:"enabled"` // Overrides Proxy field to make field nullable
+	}{}
```
Sirupsen: Why the inline type? Wouldn't it be better to define this explicitly?

README.md
```diff
@@ -429,6 +429,7 @@ All endpoints are JSON.
 
  - **GET /proxies** - List existing proxies and their toxics
  - **POST /proxies** - Create a new proxy
+ - **POST /populate** - Create or replace a list of proxies
```
Sirupsen: I think we need some more documentation on this endpoint, which parameter it takes, etc.

api_test.go
```diff
@@ -71,6 +72,203 @@ func TestCreateProxyBlankUpstream(t *testing.T) {
...
+		proxies, err := client.Proxies()
+		if err != nil {
+			t.Fatal(err)
+		} else if len(proxies) != 0 {
+			t.Fatalf("Expected no proxies to be created: %d != 0", len(proxies))
+		}
+	})
+}
+
+func TestPopulateProxyWithBadDataShouldReturnError(t *testing.T) {
```
Sirupsen: What about a test where `enabled: false` is set in the Populate endpoint?

client/client.go
```diff
@@ -130,21 +131,32 @@ func (client *Client) Proxy(name string) (*Proxy, error) {
...
+	proxies := struct {
+		Proxies []*Proxy `json:"proxies"`
+	}{}
+	request, err := json.Marshal(config)
+	if err != nil {
+		return nil, err
 	}
-	return proxies, nil
+
+	resp, err := http.Post(client.endpoint+"/populate", "application/json", bytes.NewReader(request))
```
Sirupsen: :metal:

proxy_collection.go
Sirupsen: Yeah, doing the resolving sounds good. Can be done in a follow-up, but we should create an issue for it unless you're planning to address it right away.

api.go
xthexder: This is just because of how toxics are stored internally. It's really just a json wrapper so they can be displayed properly, since the internal proxy object has them in a different format / inside a collection object.

proxy_collection.go
xthexder: Well, currently it's an issue with the client, since this logic isn't in the server yet. I'll create an issue once this is merged.

api.go
xthexder: `AddOrReplace` was added for this PR. It makes things easier and solves some concurrency / lock issues.

api.go
xthexder: I guess we could include the array index. We can't really put out the proxy name if it's the name field missing though :)

api.go
xthexder: The type is only ever used in this one function and array, so I didn't think it warranted naming. It's similar to other json marshalling code we have that uses one-time objects.

README.md
xthexder: Yep, I'll add some more info on populating proxies to the README

api_test.go
xthexder: Probably a good idea, since just enabling it all the time currently passes tests.

Sirupsen:
:ship:

PR #112, opened 2016-07-06
Updates to use replacement cli package.

claycarpenter:
Tried to build recently (from a new Go installation) and while the build succeeded, trying to list the existing proxies from the cli command yielded a deprecation warning:

```
➜  toxiproxy git:(master) ./toxiproxy-cli list
Listen		Upstream	Name	Enabled	Toxics
======================================================================
no proxies

Hint: create a proxy with `toxiproxy-cli create`
DEPRECATED Action signature.  Must be `cli.ActionFunc`.  This is an error in the application.  Please contact the distributor of this application if this is not you.  See https://github.com/urfave/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature
```

Looks like the `codegangsta/cli` package has been moved to `urface/cli`, according to that package's [README](https://github.com/urfave/cli#cli).

I updated the function signature that was causing the deprecation warning, and updated the toxiproxy-cli references to point to the updated `urface/cli` package location.

There seemed to be some issues with the latency tests. All the other tests worked fine, but the latency tests often (but not always) came in a hair too late to fit within the 10ms acceptance range. I bumped the round trip test acceptance range from 10ms to 20ms (which is just the sum of the acceptance ranges for the server and client read tests, so it seems acceptable). With that change the tests pass reliably. 

I'm very new to Go, so if I missed something here, I'd appreciate the feedback. Thanks!

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,6 +1,10 @@
 {
 	"ImportPath": "github.com/Shopify/toxiproxy",
 	"GoVersion": "go1.3.3",
+	"GodepVersion": "v74",
+	"Packages": [
+		"./..."
+	],
 	"Deps": [
 		{
 			"ImportPath": "github.com/Sirupsen/logrus",
@@ -14,6 +18,11 @@
 		{
 			"ImportPath": "github.com/gorilla/mux",
 			"Rev": "4b8fbc56f3b2400a7c7ea3dba9b3539787c486b6"
+		},
+		{
+			"ImportPath": "github.com/urfave/cli",
+			"Comment": "v1.18.0",
+			"Rev": "1efa31f08b9333f1bd4882d61f9d668a70cd902e"
 		},
 		{
 			"ImportPath": "gopkg.in/tomb.v1",
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -7,7 +7,7 @@ import (
 
 	toxiproxyServer "github.com/Shopify/toxiproxy"
 	"github.com/Shopify/toxiproxy/client"
-	"github.com/codegangsta/cli"
+	"github.com/urfave/cli"
 
 	"fmt"
 	"os"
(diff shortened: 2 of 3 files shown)


Sirupsen:
@claycarpenter great, thanks! LGTM

@jpittis ?

jpittis:
#### `cli` API Change
By always returning nil, we are complying to the new cli API but also disabling us from using the error handling feature that sparked the API change.

It seems to me like `toxiAction` should be responsible for error handling.
```go
type toxiAction func(*cli.Context, *toxiproxy.Client) error
```
And then we should be returning something like
```go
return func(c *cli.Context) error {
    return f(c, t)
}
```

Does this make sense?

This would also involve changing the type signatures of all the `toxiAction`s as well as adding a jolly `return nil` to the end of each of them.

At some point we might want to follow up with changing the error handling we currently do in our `toxiActions` to make use of the new feature.

#### Latency
I'm not sure how the 10ms was picked to begin with. Could be that the extra latency is affecting all proxies or just latency proxies. This change LGTM but we may want to open an issue looking into why it was necessary and if it's something we care about.

jpittis:
I don't see any reason why you need to address my comments above in this PR.
I'm going to merge this and create the two issues I mentioned above.

Thanks @claycarpenter 😄

jpittis:
Shouldn't this dependency change require a Godeps change?

claycarpenter:
I'm honestly too new and naive about Go to be able to answer that question about Godeps... I can only offer a shrug in response, and thanks if you can resolve that issue.

With regards to the latency tests, I should note the tests two things I saw last night:

* The tests had a somewhat random outcome. Whether all tests passed, and if not, then which ones failed, varied between test runs.
* Despite that variance, the tests always passed when I ran them against the `master` build results. 

It seems highly unlikely that the small function signature and return value change could have made enough of an impact to cause those tests to fail, but that did seem to be the case.

claycarpenter:
@jpittis Updated the Godeps manifest to reflect the changed `cli` package.

jpittis:
Sorry for taking so long to push this through @claycarpenter.
Thanks again. :)

PR #116, opened 2016-07-21
update to latest version of godeps

jpittis:
Let's cleanup godeps!

1. Update godeps to go 1.6 which changes vendor file from `Godeps/_workspace` to `vendor`.

2. We had `cli` in our `.gitignore` file which was matching against the `cli` godeps dependency. I updated the `.gitignore` to `./cli/cli` so it only refers our `cli` binary.

@Sirupsen


Sirupsen:
👍

PR #124, opened 2016-08-24
Dockerfile: upgrade to go 1.7 and build the CLI

zllak:
Upgrade Go version to 1.7, and also build the CLI.
Adapt GOPATH and PATH to launch the CLI directly from a `docker exec`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,10 @@
 FROM golang:1.7
 
-ADD . /app/src/github.com/Shopify/toxiproxy
+COPY . /go/src/github.com/Shopify/toxiproxy
 
-ENV GOPATH /app:$GOPATH
-ENV PATH $PATH:/app
-WORKDIR /app/src/github.com/Shopify/toxiproxy
-RUN go build -ldflags="-X github.com/Shopify/toxiproxy.Version=$(cat VERSION)" -o /app/toxiproxy ./cmd
-RUN go build -o /app/toxiproxy-cli ./cli
+RUN go build -ldflags="-X github.com/Shopify/toxiproxy.Version=$(cat /go/src/github.com/Shopify/toxiproxy/VERSION)" -o /go/bin/toxiproxy github.com/Shopify/toxiproxy/cmd && \
+    go build -o /go/bin/toxiproxy-cli github.com/Shopify/toxiproxy/cli
 
 EXPOSE 8474
-ENTRYPOINT ["/app/toxiproxy"]
+ENTRYPOINT ["/go/bin/toxiproxy"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,12 @@
...
+FROM golang:1.7
 
 ADD . /app/src/github.com/Shopify/toxiproxy
-RUN cd /app/src/github.com/Shopify/toxiproxy && GOPATH=/app/src/github.com/Shopify/toxiproxy/Godeps/_workspace:/app go build -ldflags="-X github.com/Shopify/toxiproxy.Version $(cat VERSION)" -o /app/toxiproxy ./cmd
+
+ENV GOPATH /app:$GOPATH
+ENV PATH $PATH:/app
+WORKDIR /app/src/github.com/Shopify/toxiproxy
+RUN go build -ldflags="-X github.com/Shopify/toxiproxy.Version=$(cat VERSION)" -o /app/toxiproxy ./cmd
+RUN go build -o /app/toxiproxy-cli ./cli
```
Sirupsen: Do we need to build 5 extra layers for this?

Good call on building the CI.

Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM golang:1.4
+FROM golang:1.7
 
 ADD . /app/src/github.com/Shopify/toxiproxy
-RUN cd /app/src/github.com/Shopify/toxiproxy && GOPATH=/app/src/github.com/Shopify/toxiproxy/Godeps/_workspace:/app go build -ldflags="-X github.com/Shopify/toxiproxy.Version $(cat VERSION)" -o /app/toxiproxy ./cmd
+
+ENV GOPATH /app:$GOPATH
```
xthexder: Will this actually use the Godeps workspace? I'm not sure what the default $GOPATH is here.

Dockerfile
zllak: Hey Sirupsen (love logrus man :D),

You're right, too much layers, not necessary.
I've updated the Dockerfile.

Dockerfile
zllak: You're right, it's not using Godeps workspace, but there is a vendor directory already, so Godeps might not be very useful.
It should even be possible to completely remove Godeps from toxiproxy! :)

xthexder: Ah interesting. Looks like godeps updated to use the vendor directory instead. Alright 👍

jpittis: Given that vendor is stable in 1.7 now, I think we should be switching away from godeps.

Sirupsen:
LGTM

Issue #122, opened 2016-08-22
Data limit toxic suggestion


matez:
I need to simulate situation when client sends too much data and server close the connection. Would you accept PR with simple toxic which would do such thing? ... limit how much data can be transferred during single connection but let pass everything before threshold is reached.

Sirupsen:
Yeah, that sounds like a great idea!

xthexder:
Toxiproxy 2.1.0 has been released.

PR #125, opened 2016-08-24
Add toxic limiting size of transmitted data

matez:
As mentioned in #122 here is PR ... sort of first experience with golang.

### DIFF
--- a/toxics/limit_data.go
+++ b/toxics/limit_data.go
@@ -1,0 +1,47 @@
+package toxics
+
+import "github.com/Shopify/toxiproxy/stream"
+
+// LimitDataToxic has limit in bytes
+type LimitDataToxic struct {
+	Bytes int64 `json:"bytes"`
+}
+
+func (t *LimitDataToxic) Pipe(stub *ToxicStub) {
+	var bytesRemaining = t.Bytes
+
+	for {
+		select {
+		case <-stub.Interrupt:
+			return
+		case c := <-stub.Input:
+			if c == nil {
+				stub.Close()
+				return
+			}
+
+			chunk := c
+
+			if bytesRemaining < int64(len(c.Data)) {
+				chunk = &stream.StreamChunk{
+					Timestamp: c.Timestamp,
+					Data:      c.Data[0:bytesRemaining],
+				}
+			}
+
+			if len(chunk.Data) > 0 {
+				stub.Output <- chunk
+				bytesRemaining -= int64(len(chunk.Data))
+			}
+
+			if bytesRemaining <= 0 {
+				stub.Close()
+				return
+			}
+		}
+	}
+}
+
+func init() {
+	Register("limit_data", new(LimitDataToxic))
+}
--- a/toxics/limit_data.go
+++ b/toxics/limit_data.go
@@ -4,11 +4,12 @@ import "github.com/Shopify/toxiproxy/stream"
 
 // LimitDataToxic has limit in bytes
 type LimitDataToxic struct {
-	Bytes int64 `json:"bytes"`
+	Bytes            int64 `json:"bytes"`
+	bytesTransmitted int64
 }
 
 func (t *LimitDataToxic) Pipe(stub *ToxicStub) {
-	var bytesRemaining = t.Bytes
+	var bytesRemaining = t.Bytes - t.bytesTransmitted
 
 	for {
 		select {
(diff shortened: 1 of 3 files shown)


toxics/limit_data.go
```diff
@@ -0,0 +1,47 @@
...
+
+import "github.com/Shopify/toxiproxy/stream"
+
+// LimitDataToxic has limit in bytes
+type LimitDataToxic struct {
+	Bytes int64 `json:"bytes"`
+}
+
+func (t *LimitDataToxic) Pipe(stub *ToxicStub) {
+	var bytesRemaining = t.Bytes
```
xthexder: Unfortunately the way toxics are loaded and interrupted, this value won't be persistent.
There's several cases were a toxic is interrupted and the `Pipe()` function is restarted.

I think the solution to this is to put a private `bytesTransmitted` variable in the toxic struct. It will default to 0 when the toxic is created, and you can increment it when you send bytes.

I'll have to check through the rest of the toxiproxy code to make sure there's no places were this value would get reset.

toxics/limit_data_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/Shopify/toxiproxy/stream"
+	"github.com/Shopify/toxiproxy/toxics"
+)
+
+func cmpBuffers(bufa []byte, bufb []byte) bool {
```
xthexder: This exists as a built in go library: https://golang.org/pkg/bytes/#Equal

toxics/limit_data_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return true
+}
+
+func buffer(size int) []byte {
+	buf := make([]byte, size)
+
+	for i := 0; i < size; i++ {
+		buf[i] = byte(rand.Int())
```
xthexder: You could simplify this loop a bit with `rand.Read(buf)`

toxics/limit_data_test.go
```diff
@@ -0,0 +1,136 @@
...
+	return buf
+}
+
+func check(t *testing.T, toxic *toxics.LimitDataToxic, chunks [][]byte, expectedChunks [][]byte) {
+	input := make(chan *stream.StreamChunk)
+	output := make(chan *stream.StreamChunk, 100)
+	stub := toxics.NewToxicStub(input, output)
+
+	go func() {
+		toxic.Pipe(stub)
```
xthexder: This can actually just be `go toxic.Pipe(stub)`

toxics/limit_data_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for _, buf := range chunks {
+		input <- &stream.StreamChunk{Data: buf}
+	}
+
+	for _, expected := range expectedChunks {
+		chunk := <-output
+
+		if !cmpBuffers(chunk.Data, expected) {
+			t.Fail()
```
xthexder: Using `t.Error("message")` would make any test failures much more descriptive.
Something like "Chunk data did not match expected" and "Too many chunks returned by toxic"

toxics/limit_data_test.go
```diff
@@ -0,0 +1,136 @@
...
+	toxic := &toxics.LimitDataToxic{Bytes: 100}
+
+	buf := buffer(100)
+	check(t, toxic, [][]byte{buf}, [][]byte{buf})
+}
+
+func TestLimitDataToxicChunkBiggerThanLimit(t *testing.T) {
+	toxic := &toxics.LimitDataToxic{Bytes: 100}
+
+	buf := buffer(100)
```
xthexder: Shouldn't this be bigger than 100?

toxics/limit_data.go
xthexder: Alright, I confirmed this, and any fields in the toxic struct will be persisted until the toxic is removed.

toxics/limit_data_test.go
```diff
@@ -0,0 +1,136 @@
...
+			t.Fail()
+		}
+	}
+
+	if len(output) != 0 {
+		t.Fail()
+	}
+}
+
+func TestLimitDataToxicMayBeInterrupted(t *testing.T) {
```
xthexder: This could maybe be adapted to test that the limit still works after interrupting and restarting a toxic.

xthexder:
This looks pretty good. I've been a bit nit-picky, and theirs some weird internal behaviour that I mentioned, that's an easy fix.

Thanks for the contribution!

matez:
Thank you for your comments. Very helpful. I'll make appropriate changes tomorrow.

PR #127, opened 2016-09-01
Add limit data toxic

xthexder:
Adds https://github.com/Shopify/toxiproxy/pull/125 with some fixes
I made some simplifications to `toxic_collection.go` that don't change anything other than make the code look nicer (no more duplicated `c.chain` and `c.toxics`)

Stateful toxics are added to allow for per-connection toxic state to be stored.
This fixes the issue of `bytesTransmitted` in the limit_data toxic being shared between connections.

@Sirupsen @pushrax 
cc @matez

### DIFF
--- a/CREATING_TOXICS.md
+++ b/CREATING_TOXICS.md
@@ -115,6 +115,24 @@ The unit used by `GetBufferSize()` is `StreamChunk`s. Chunks are generally anywh
 1 byte, up to 32KB, so keep this in mind when thinking about how much buffering you need,
 and how much memory you are comfortable with using.
 
+## Stateful toxics
+
+If a toxic needs to store extra information for a connection such as the number of bytes
+transferred (See `limit_data` toxic), a state object can be created by implementing the
+`StatefulToxic` interface. This interface defines the `NewState()` function that can create
+a new state object with default values set.
+
+When a stateful toxic is created, the state object will be stored on the `ToxicStub` and
+can be accessed from `toxic.Pipe()`:
+
+```go
+state := stub.State.(*ExampleToxicState)
+```
+
+If necessary, some global state can be stored in the toxic struct, which will not be
+instanced per-connection. These fields cannot have a custom default value set and will
+not be thread-safe, so proper locking or atomic operations will need to be used.
+
 ## Using `io.Reader` and `io.Writer`
 
 If your toxic involves modifying the data going through a proxy, you can use the `ChanReader`
--- a/link_test.go
+++ b/link_test.go
@@ -207,7 +207,6 @@ func TestStateCreated(t *testing.T) {
 	go link.stubs[0].Run(collection.chain[stream.Downstream][0])
 	collection.links["test"] = link
 
-	// Add stubs
 	collection.chainAddToxic(&toxics.ToxicWrapper{
 		Toxic:     new(toxics.LimitDataToxic),
 		Type:      "limit_data",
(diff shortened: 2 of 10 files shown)


xthexder:
Also @jpittis for possible review

CREATING_TOXICS.md
```diff
@@ -115,6 +115,24 @@ The unit used by `GetBufferSize()` is `StreamChunk`s. Chunks are generally anywh
 1 byte, up to 32KB, so keep this in mind when thinking about how much buffering you need,
 and how much memory you are comfortable with using.
 
+## Stateful toxics
+
+If a toxic needs to store extra information for a connection such as the number of bytes
+transferred (See `limit_data` toxic), a state object can be created by implementing the
+`StatefulToxic` interface. This interface defines the `NewState()` function that can create
```
Sirupsen: Why not make `StatefulToxic` the default?

xthexder: This way the general / common case is simpler. Most toxics are stateless, so they would all have a `NewState()` function that just returns `nil`. This way the function is optional. It operates the same way as `BufferedToxic`, which defaults to a buffer size of 0.

Sirupsen: I suppose since it's implicit, it's fine.

CREATING_TOXICS.md
```diff
@@ -115,6 +115,24 @@ The unit used by `GetBufferSize()` is `StreamChunk`s. Chunks are generally anywh
...
+When a stateful toxic is created, the state object will be stored on the `ToxicStub` and
+can be accessed from `toxic.Pipe()`:
+
+```go
+state := stub.State.(*ExampleToxicState)
+```
+
+If necessary, some global state can be stored in the toxic struct, which will not be
+instanced per-connection. These fields cannot have a custom default value set and will
+not be thread-safe, so proper locking or atomic operations will need to be used.
```
Sirupsen: 👍 for including this

link_test.go
```diff
@@ -200,3 +200,21 @@ func TestToxicity(t *testing.T) {
...
 	}
 }
+
+func TestStateCreated(t *testing.T) {
+	collection := NewToxicCollection(nil)
+	link := NewToxicLink(nil, collection, stream.Downstream)
+	go link.stubs[0].Run(collection.chain[stream.Downstream][0])
+	collection.links["test"] = link
+
+	// Add stubs
```
Sirupsen: Remove comment

toxic_collection.go
```diff
@@ -46,36 +46,32 @@ func (c *ToxicCollection) ResetToxics() {
...
 
-	for dir := range c.toxics {
-		for _, toxic := range c.toxics[dir] {
-			if toxic.Name == name {
-				return toxic
-			}
-		}
-	}
-	return nil
+	return c.findToxicByName(name)
```
Sirupsen: This simplification is great.

toxics/limit_data.go
```diff
@@ -0,0 +1,60 @@
...
+	Bytes int64 `json:"bytes"`
+}
+
+type LimitDataToxicState struct {
+	bytesTransmitted int64
+}
+
+func (t *LimitDataToxic) Pipe(stub *ToxicStub) {
+	state := stub.State.(*LimitDataToxicState)
+	var bytesRemaining = t.Bytes - state.bytesTransmitted
```
Sirupsen: Why would any bytes have been transmitted at this point?

Sirupsen:
Couple of questions but overall looks good!

toxics/limit_data.go
xthexder: Toxics can be interrupted at any point and restarted, so some bytes may have been transmitted, the toxic is updated, and then `Pipe()` is restarted, but with the same state as before.

Issue #128, opened 2016-09-01
cli: Allow host to be specified


dylanahsmith:
At Shopify we use a VM to run services that a project depends on, including toxiproxy.  Unfortunately, toxiprox-cli is hard coded to use localhost.  That means that toxiproxy-cli can't be used from outside of the VM

I think there should be a global option to specify the host.  E.g. `toxiproxy-cli --host=example.com list`

xthexder:
This should be easy to add, I'll throw together a PR.

xthexder:
Toxiproxy 2.1.0 has been released.

PR #129, opened 2016-09-02
Change cli to allow remote toxiproxy hosts

xthexder:
Adds `--host / -h` to specify the toxiproxy host to connect to.

For https://github.com/Shopify/toxiproxy/issues/128

@Sirupsen 
cc @dylanahsmith


Sirupsen:
❤️  A changelog entry and then 🚢

Issue #50, opened 2015-05-05
Protocol aware toxics


Sirupsen:
We could build protocol aware toxics, e.g.

* **HTTP**. Send extremely large headers, slow headers, slow body but fast headers, ..
* **Redis**. Slow writes and fast reads, out of memory conditions (reads but no writes), ..

dwradcliffe:
This would be neat. I'd like to be able to return HTTP 404 and test that.

Sirupsen:
@dwradcliffe it's completely possible to build already BTW if you really need it. We just haven't. Thanks for the input on a use-case :-)

dwradcliffe:
My specific use-case is for elasticsearch. ES often returns non-200 responses when things are not quite right or when the request was routed to the wrong place.

xthexder:
I wrote a simple HTTP toxic while writing the docs for 2.0
You can find it here: https://github.com/xthexder/toxic-example/blob/master/http.go

If you're up for it, you could modify it for ES, and compile it in to the 2.0-dev branch.
Hopefully I'll have some time in the next week or so to release 2.0, but I've got a fair amount of other work to do at the moment.

Sirupsen:
@xthexder how about taking the HTTP toxic into core?

xthexder:
The toxic needs some work to generalize the usage, since it needs to be controllable via json, but it's definitely something we can put in core.
It will also need a matching request http toxic, since this one is only for http responses.

xthexder:
This might be interesting to bring up again. I'm not sure of all the use-cases for this, but a couple things I can think of to put as options are:

 - Change status code
 - Add/remove headers
 - Set response body
 - Toxicity per-request vs per-connection

I can work on implementing some of these, but I'd like to hear some others opinions.

For some of the things I'm thinking of, there's the problem that downstream toxics can't read any information from the upstream / request, so things can't be filtered based on URL for example.

Sirupsen:
The status code change one would be interesting for sure, adding removing random headers, etc., would all be awesome.

> there's the problem that downstream toxics can't read any information from the upstream / request

Ugh, this is a great point. How hard would it be to add this information though?

xthexder:
I'm currently looking at it, and I think it's possible to add, it just might be a little messy with matching corresponding upstream/downstream toxics together.
The question is mostly if it will be transparent to the user/API or not.

aebicalho:
[...]

PR #132, opened 2016-09-14
Add Bidirectional Toxics

xthexder:
This is in preparation for adding an HTTP toxic as well as enabling other protocol-aware toxics.

This PR does a few things:
 - Keeps track of paired links so that for example an `downstream` link can reference an `upstream` link.
 - Adds the idea of paired toxics, allowing for an `upstream` and `downstream` toxic to share a state object.
 - Hides toxics with a blank name so that internal toxics aren't visible (A bidirectional toxic actually creates 2 toxics).
 - Handles toxicity so that paired toxics are always in the same state
 - Lots of tests


toxic_collection.go
```diff
@@ -96,20 +96,31 @@ func (c *ToxicCollection) AddToxicJson(data io.Reader) (*toxics.ToxicWrapper, er
...
+		switch strings.ToLower(wrapper.Stream) {
+		case "downstream":
+			wrapper.Direction = stream.Downstream
+		case "upstream":
+			wrapper.Direction = stream.Upstream
+		default:
+			return nil, ErrInvalidStream
+		}
+	} else {
+		wrapper.Stream = "both"
```
xthexder: I'm not sure what the behaviour should be here since there's only one option for bidirectional toxics.
Right now `stream` can be set to anything and it will just be changed to `both` by the server.

xthexder:
Updates to `CHANGELOG.md` and `CREATING_TOXICS.md` to come.

xthexder:
I think this should be ready for review now, @Sirupsen 
Feel free to add some other people to review, I'm not sure who else is interested in Go + Toxiproxy right now.

The WIP http toxic is here: https://github.com/Shopify/toxiproxy/blob/http-toxic/toxics/http.go
And another example of an implemented bidirectional toxic is in the tests: https://github.com/Shopify/toxiproxy/blob/bidirectional-toxics/toxics/bidirectional_test.go#L17

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 2.1.0 (Unreleased)
 
+* Add bidirectional toxics #132
```
Sirupsen: Could we add a little bit more of a comment here on why this is useful? What it allows us to do?

toxics/toxic.go
```diff
@@ -39,21 +40,30 @@ type StatefulToxic interface {
 	NewState() interface{}
 }
 
+// Bidirectional toxics operate on both TCP streams and allow state to be shared.
+// PipeRequest() will oparate on the upstream, while Pipe() will operate on the downstream.
+type BidirectionalToxic interface {
+	// Defines the packet flow through an upstream ToxicStub. Operates the same as Pipe().
+	PipeRequest(*ToxicStub)
```
Sirupsen: Should we call this `PipeUpstream` instead to keep the terminology consistent?

CREATING_TOXICS.md
```diff
@@ -133,6 +142,78 @@ If necessary, some global state can be stored in the toxic struct, which will no
...
+this is a problem.
+
+Bidirectional toxics allow state to be shared for the `upstream` and `downstream` pipes in a single
+toxic implementation. They also ensure direction-specific code is always run on the correct pipe
+(a toxic that only works on the `upstream` can't be added to the `downstream`).
+
+Creating a bidirectional toxic is done by implementing a second `Pipe()` function called `PipeRequest()`.
+The implementation is same as a regular toxic, and can be paired with other types such as a stateful toxic.
+
+One use case of a bidirectional toxic is to mock out the backend server entirely, which is shown below:
```
Sirupsen: I think it'd be more clear in people's mind if we elaborated this example a little bit to a use-case, e.g. what if we did a simple implementation of the Redis protocol (which is super simple) that fails writes of keys directly? Something like..

```go
case c := <-stub.Request:
  if c == nil {
   stub.Close()
   return
  }
  if strings.HasPrefix(string(c), "SET") {
    c <- []byte() # failed to write
  } else {
    c <- []byte("OK")
  }
```

I'm sure this example doesn't work, but it illustrates it. An alternative would be an HTTP server, but that may be a little bit harder to make as a concise example because the protocol is more complex and AFAIK the Go STDLIB doesn't expose the HTTP parser nicely (or maybe it does?) in which case one that sends 5xx in some cases would be cool.

I'm assuming this will work out of the box with toxicity? In which case perhaps above instead of doing anything fancy, you can just return a 500 HTTP response in the example in `<toxicity>` of cases.

link.go
```diff
@@ -67,7 +68,12 @@ func (link *ToxicLink) Start(name string, source io.Reader, dest io.WriteCloser)
 	}()
 	for i, toxic := range link.toxics.chain[link.direction] {
 		if stateful, ok := toxic.Toxic.(toxics.StatefulToxic); ok {
-			link.stubs[i].State = stateful.NewState()
+			if toxic.PairedToxic == nil || link.pairedLink.stubs[toxic.PairedToxic.Index].State == nil {
```
Sirupsen: I'm not a huge fan of this added complexity, but since BidirectionalToxics override the entire response I don't see an easy way out. Perhaps we should emphasize more heavily that they're fundamentally different toxics because they don't parse anything to the server...

Another approach we could take is that `PipeRequest` can choose whether it passes directly to the downstream toxic or to the backend, in that case the default template for toxics is just to pass the data as is to the upstream. This also allows for some new interesting toxics where you manipulate with the data that's being sent to the backend, e.g. wrong key names.

Could one interface in that case work for all use-cases? Even if it's a bit more verbose. Realistically not that many toxics will be created by third-parties for now, so I think simplifying the core of Toxiproxy for more verbose toxics (that are more powerful?) isn't a bad trade-off to make.

toxics/toxic.go
xthexder: Yeah, probably a good idea. I was originally writing this as part of the HTTP toxic, so that's where the naming came from.

CREATING_TOXICS.md
xthexder: A redis example could be interesting, I'll see if I can throw something together.
Go does expose it's HTTP request/response parsing, but it requires using `io.Reader` / `io.Writer`, so it's maybe a little complicated for an example. (You can see it here https://github.com/Shopify/toxiproxy/blob/http-toxic/toxics/http.go)

link.go
xthexder: I guess my documentation could still use some work. All of these things are possible with the current interface.
A bidirectional toxic internally is essentially just an upstream and downstream toxic paired together, with the option of sharing a user-defined state object.
They can be completely independent and work like any other toxic, or they can do fancy things with the state to share data or even completely bypass the backend server.

I originally tried to hide this implementation detail in the docs, but maybe that wasn't such a great idea and makes things more confusing.

CREATING_TOXICS.md
Sirupsen: Yeah, that might be too complicated for now.

sirupsen:
@jpittis is this something you'd be interested in reviving? Protocol aware toxics would be awesome (https://github.com/Shopify/toxiproxy/issues/50).

mmollaverdi:
+1. This seems very useful.

mgkeen:
I was hacking together a MS SQL TDS aware toxic to cut off connections when given batch or RPC calls are sent upstream. If this ever gets anywhere then give me a nudge and i will finish it off 😄

Issue #113, opened 2016-07-06
better CLI error handling


jpittis:
https://github.com/Shopify/toxiproxy/pull/112 updates our CLI library to a newer version with better error handling. We should figure out how this error handling works and change `toxiAction` to support the new error handling if it's a cleaner solution.

`toxiAction` should be updated to look something like this.

``` go
type toxiAction func(*cli.Context, *toxiproxy.Client) error

return func(c *cli.Context) error {
    return f(c, t)
}
```

And then all the action error handling should be ported to the new API.

xthexder:
Toxiproxy 2.1.0 has been released.

PR #133, opened 2016-09-19
Scriptable CLI

xthexder:
The cli has proven itself to be very useful for quickly testing things, but until now it hasn't been very useful as part of scripts and automation.
This PR adds some checks to see if a real TTY is being used, and changes the output format if the cli is used as part of a pipe:

## Example:
```
∙ ./toxiproxy-cli list
Name   			Listen 		Upstream       		Enabled		Toxics
======================================================================================
4      			[::]:8083      	google.com:80  		enabled		None
google 			[::]:8080      	google.com:80  		enabled		3
google2			[::]:8081      	google.com:80  		enabled		None
google3longggggg       	[::]:8082      	google.com:80  		enabled		None

Hint: inspect toxics with `toxiproxy-cli inspect <proxyName>`

∙ ./toxiproxy-cli list | cat
4      	[::]:8083      	google.com:80  	enabled	0
google 	[::]:8080      	google.com:80  	enabled	3
google2	[::]:8081      	google.com:80  	enabled	0
google3longggggg       	[::]:8082      	google.com:80  	enabled	0
```

This allows for some interesting usage from bash:
```
∙ ./toxiproxy-cli ls | cut -f 1 | xargs -n1 ./toxiproxy-cli toggle
Proxy 4 is now disabled
Proxy google is now disabled
Proxy google2 is now disabled
Proxy google3longggggg is now disabled

∙ ./toxiproxy-cli i google | grep upstream | cut -f 1 | xargs -n1 ./toxiproxy-cli toxic update google -tox 0.5 -n
Updated toxic 'latency_upstream' on proxy 'google'
Updated toxic 'bandwidth_upstream' on proxy 'google'
```

## Other Changes
While writing this I noticed some problems with how toxic attributes were parsed (mainly only supporting integer values), so I've updated that as well to the following format:
`./toxiproxy-cli toxic create google -t latency -a latency=500 -a jitter=10`

This format supports integers, floats, and string fields.

This PR also updates the cli to use the new error handling api: https://github.com/Shopify/toxiproxy/issues/113

@Sirupsen @pushrax @jpittis

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
@@ -8,21 +10,38 @@ import (
 	toxiproxyServer "github.com/Shopify/toxiproxy"
 	"github.com/Shopify/toxiproxy/client"
 	"github.com/urfave/cli"
-
-	"fmt"
-	"os"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 const (
-	redColor    = "\x1b[31m"
-	greenColor  = "\x1b[32m"
-	yellowColor = "\x1b[33m"
-	blueColor   = "\x1b[34m"
-	cyanColor   = "\x1b[36m"
-	purpleColor = "\x1b[35m"
-	grayColor   = "\x1b[37m"
-	noColor     = "\x1b[0m"
+	RED = iota
+	GREEN
+	YELLOW
+	BLUE
+	CYAN
+	PURPLE
+	GRAY
+	NONE
 )
+
+var colors = map[int]string{
+	RED:    "\x1b[31m",
+	GREEN:  "\x1b[32m",
+	YELLOW: "\x1b[33m",
+	BLUE:   "\x1b[34m",
+	CYAN:   "\x1b[36m",
+	PURPLE: "\x1b[35m",
+	GRAY:   "\x1b[37m",
+	NONE:   "\x1b[0m",
+}
+
+func color(color int) string {
+	if isTTY {
+		return colors[color]
+	} else {
+		return ""
+	}
+}
 
 var toxicDescription = `
 	Default Toxics:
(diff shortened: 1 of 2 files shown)


cli/cli.go
```diff
@@ -1,29 +1,48 @@
...
+	GREEN
+	YELLOW
+	BLUE
+	CYAN
+	PURPLE
+	GRAY
+	NONE
 )
 
+var colors = map[int]string{
```
jpittis: Are there other advantages to having this hash map besides tty? Why not just have a
```go
func color(color string) string {
    if isTTY {
        return color
    } else {
        return ""
    }
}
```

and then use `color(BLUE)` with string constant `BLUE` rather than enum.

cli/cli.go
```diff
@@ -222,249 +242,293 @@ func list(c *cli.Context, t *toxiproxy.Client) {
...
-	for _, keyValue := range keyValues {
-		kv := strings.SplitN(keyValue, "=", 2)
-		if len(kv) != 2 {
-			fatalf("Attributes should be in format key1=value1,key2=value2\n")
+	args := c.StringSlice(name)
+
+	for _, raw := range args {
+		kv := strings.SplitN(raw, "=", 2)
+		if len(kv) < 2 {
+			continue
```
jpittis: Why should this not return an error rather than `continue`?

cli/cli.go
```diff
@@ -496,40 +560,59 @@ func sortedAttributes(attrs toxiproxy.Attributes) attributeList {
...
+
+func errorf(m string, args ...interface{}) error {
+	return cli.NewExitError(fmt.Sprintf(m, args...), 1)
 }
 
-func fatalf(m string, args ...interface{}) {
-	fmt.Printf(m, args...)
-	os.Exit(1)
+func printWidth(col int, m string, numTabs int) {
+	if !isTTY {
```
jpittis: Nitpick: Everywhere else we do `if isTTY` not `if !isTTY`.

cli/cli.go
xthexder: I guess it's just a question of if something like `-a foo` should error or just be ignored. Right now it's just ignored, which seems fine to me, it will effectively just set a blank / 0 value in the toxic this way.

PR #134, opened 2016-10-04
Update ChanReader / ChanWriter for use inside toxics

xthexder:
While working on the bidirectional toxics PRs I discovered some problems with how `ChanReader` worked when used ephemerally inside a toxic.
These turned out to be non-trivial to fix, with the end result being this PR.

The problem with `ChanReader` is that in some cases it may buffer data. This means if you ever stop using the reader and switch to another, there's a possibility of losing data.
Further problems arise when `ChanReader` is read using other readers such as `bufio.Reader` as required by functions such as `http.ReadRequest()`.

These problems are solved in 3 ways:
 1. Use a non-ephemeral reader object on the `ToxicStub` instead of creating it every time `Pipe()` is called.
 2. Create a `TransactionalReader` that can be rolled back to specific points so that `bufio.Reader` can effectively be un-read.
 3. Allow all buffers inside the reader to be explicitly flushed at any point so that no data is lost.

Example `TransactionalReader` usage:
```go
// Input is "hello world"
txnReader := stream.NewTransactionalReader(input)

{ // This section is ephemeral (and would be similar to usage in `Pipe()`)
    bufReader := bufio.NewReader(txnReader)
    msg, _ := bufReader.ReadString(' ')
    // msg == "hello "
    // bufReader has "world" buffered (bufReader.Buffered() == 5)

    // Mark everything up to "hello " as read.
    txnReader.Checkpoint(-bufReader.Buffered())
    // Rollback txnReader to the checkpoint, we never read "world"
    txnReader.Rollback()
    // bufReader can now be thrown away
}

// Any further reads from txnReader will start from the checkpoint
buf := make([]byte, 5)
txnReader.Read(buf)
// buf == "world"
```

Actual usage can be found in my WIP redis toxic here: https://github.com/Shopify/toxiproxy/blob/redis-wip/toxics/redis.go

The documentation for this is pretty tricky to write without actual code examples, so I'll likely leave the detailed docs until we have at least 1 protocol-aware toxic to point at.

@Sirupsen @jpittis / anyone else interested in Golang


xthexder:
Went through this again and cleaned some things up now that it's not fresh in my mind.
Should be good for a review @sirupsen

CREATING_TOXICS.md
```diff
@@ -145,19 +145,16 @@ An implementation of the noop toxic above using the stream package would look so
...
         if err == stream.ErrInterrupted {
-            writer.Write(buf[:n])
             return
         } else if err == io.EOF {
             stub.Close()
             return
         }
-        writer.Write(buf[:n])
+        stub.Writer.Write(buf[:n])
+        stub.Reader.Checkpoint(0)
```
jpittis: Why `Checkpoint(0)`? Should it not be `Checkpoint(-stub.Reader.Buffered())`?

stream/io_chan.go
```diff
@@ -106,3 +115,104 @@ func (c *ChanReader) Read(out []byte) (int, error) {
...
+type TransactionalReader struct {
+	buffer    *bytes.Buffer
+	bufReader *bytes.Reader
+	reader    *ChanReader
+	tee       io.Reader
+}
+
+func NewTransactionalReader(input <-chan *StreamChunk) *TransactionalReader {
+	t := &TransactionalReader{
+		buffer: bytes.NewBuffer(make([]byte, 0, 32*1024)),
```
jpittis: Did you choose 32K sized buffer for a specific reason?

stream/io_chan.go
```diff
@@ -106,3 +115,104 @@ func (c *ChanReader) Read(out []byte) (int, error) {
...
+// If offset is negative, the checkpoint will be set N bytes before the current position.
+// If the offset is positive, the checkpoint will be set N bytes after the previous checkpoint.
+// An offset of 0 will set the checkpoint to the current position.
+func (t *TransactionalReader) Checkpoint(offset int) {
+	current := t.buffer.Len()
+	if t.bufReader != nil {
+		current = int(t.bufReader.Size()) - t.bufReader.Len()
+	}
+
+	n := current
```
jpittis: Initializing `n` to `current` seems to be pointless here? Why not `var n int`?

CREATING_TOXICS.md
xthexder: The example where I used `.Buffered()` was for a `bufio.Reader` attached to the `TransactionalReader`.
In this case, the amount of buffered data is always 0. Since everything read is being used, we can set the checkpoint to the last read byte, rather than the `(last read) - (amount buffered by bufio.Reader)`

stream/io_chan.go
xthexder: You're right, looks like this is redundant. I'll see if I can refactor this function and make it a little simpler.

xthexder: Actually found another bug while looking at this. if `bufReader != nil`, `t.buffer.Reset()` shouldn't be called unless the offset is big enough.
I fixed this, and added a test for it.

stream/io_chan.go
```diff
@@ -106,3 +115,102 @@ func (c *ChanReader) Read(out []byte) (int, error) {
...
+
+// Sets a checkpoint in the reader. A call to Rollback() will begin reading from this point.
+// If offset is negative, the checkpoint will be set N bytes before the current position.
+// If the offset is positive, the checkpoint will be set N bytes after the previous checkpoint.
+// An offset of 0 will set the checkpoint to the current position.
+func (t *TransactionalReader) Checkpoint(offset int) {
+	current := t.buffer.Len()
+	if t.bufReader != nil {
+		current = int(t.bufReader.Size()) - t.bufReader.Len()
+	}
```
jpittis: So `current` represents how much of the buffer we've already read.
- `0` will say... okay we've read everything! Lock it in!
- `-5` will say we've read everything except for the last 5 bytes. Let's start reading again from the start of those last 5.
- `5` will say we successfully read 5 bytes but not the rest. Let's start reading from the rest.
- `t.buffer.Len()` will do the same thing is `0`.

Okay I found this confusing but LGTM

stream/io_chan.go
```diff
@@ -106,3 +115,102 @@ func (c *ChanReader) Read(out []byte) (int, error) {
...
+	if t.bufReader != nil {
+		n = t.bufReader.Len()
+	}
+	buf := make([]byte, n+len(t.reader.buffer))
+	if n > 0 {
+		t.bufReader.Read(buf[:n])
+	}
+	if len(buf[n:]) > 0 {
+		t.reader.Read(buf[n:])
+	}
```
jpittis: Is this correct? Why are we trying to flush the reader? It's not buffered is it? I'll look at this again later and probably understand right away.

PR #153, opened 2016-12-06
Adding static configuration to toxi proxy

delta-bravo:
Currently the proxies are volatile and their config vanishes after restart. In our environment we need to be able to specify a few upstream service endpoints that never change and have to be available from toxiproxy startup.

The purpose of this request is to add static proxies configuration as seen in config.json. Sorry about empty lines in api.go and proxy_collection.go

THIS SOFTWARE CONTRIBUTION IS PROVIDED ON BEHALF OF SKY PLC. BY THE CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED

### DIFF
--- a/api.go
+++ b/api.go
@@ -46,6 +46,9 @@ func (server *ApiServer) Listen(host string, port string) {
 		"version": Version,
 	}).Info("API HTTP server starting")
 
+
+
+
 	err := http.ListenAndServe(net.JoinHostPort(host, port), nil)
 	if err != nil {
 		log.Fatal("ListenAndServe: ", err)
--- a/cmd/toxiproxy.go
+++ b/cmd/toxiproxy.go
@@ -6,10 +6,24 @@ import (
 	"time"
 
 	"github.com/Shopify/toxiproxy"
+
+	"io/ioutil"
+	"github.com/Sirupsen/logrus"
+	"encoding/json"
 )
 
 var host string
 var port string
+
+type ProxyConfig struct {
+	Name string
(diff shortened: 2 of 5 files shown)


xthexder:
This is basically duplicating the functionality of the `populate` function in most clients (and is also a built-in endpoint in the next version)

Normally `populate` is just called when your app/tests boot, but a cli argument for the server would also be useful.
I'd suggest adding an argument like `-config <config.json>` that will do the same as https://github.com/Shopify/toxiproxy/blob/master/api.go#L134
Keeping a consistent format between the json file and the endpoint would also be good.

delta-bravo:
Thanks for the comment, I will change this shortly.

sirupsen:
I am +1 to a `-config` that uses `populate()` behind the scenes. We haven't seen a need for this in our environment, but I could definitely see how this could be useful in others.

xthexder:
This functionality will be released with toxiproxy `2.1.0`, and should solve your use-case.

delta-bravo:
@xthexder is any input required from my side then?

xthexder:
@delta-bravo You should be good to go. I'm currently in the process of pushing the 2.1.0 release.

delta-bravo:
Thanks for your help!

PR #154, opened 2016-12-07
Add -config option for toxiproxy startup

xthexder:
I'd like to get a 2.1.0 released soon, since there's a lot of unreleased features piling up.
I put together the functionality described in https://github.com/Shopify/toxiproxy/pull/153 in this PR so we can get it included in 2.1.0

Example:
```
[
  {
    "name": "redis",
    "upstream": "localhost:6379",
    "listen": "localhost:6380"
  }
]
```

```
$ ./toxiproxy-server -config test.json
INFO[0000] Started proxy                                 name=redis proxy=127.0.0.1:6380 upstream=localhost:6379
INFO[0000] Populated proxies from file                   config=test.json proxies=1
INFO[0000] API HTTP server starting                      host=localhost port=8474 version=git-8e40aad
INFO[0002] Accepted client                               client=127.0.0.1:57170 name=redis proxy=127.0.0.1:6380 upstream=localhost:6379
```

```
$ redis-cli -p 6380
127.0.0.1:6380> SET foo bar
OK
127.0.0.1:6380> GET foo
"bar"
```

@sirupsen @delta-bravo

### DIFF
--- a/cmd/toxiproxy.go
+++ b/cmd/toxiproxy.go
@@ -3,11 +3,9 @@ package main
 import (
 	"flag"
 	"math/rand"
-	"os"
 	"time"
 
 	"github.com/Shopify/toxiproxy"
-	"github.com/Sirupsen/logrus"
 )
 
 var host string
@@ -26,26 +24,7 @@ func init() {
 func main() {
 	server := toxiproxy.NewServer()
 	if len(config) > 0 {
-		file, err := os.Open(config)
-		if err != nil {
-			logrus.WithFields(logrus.Fields{
-				"config": config,
-				"error":  err,
-			}).Error("Error reading config file")
-		} else {
-			proxies, err := server.Collection.PopulateJson(file)
-			if err != nil {
-				logrus.WithFields(logrus.Fields{
-					"config": config,
-					"error":  err,
-				}).Error("Failed to populate proxies from file")
-			} else {
-				logrus.WithFields(logrus.Fields{
-					"config":  config,
-					"proxies": len(proxies),
-				}).Info("Populated proxies from file")
-			}
-		}
+		server.PopulateConfig(config)
 	}
 	server.Listen(host, port)
 }
--- a/cmd/toxiproxy.go
+++ b/cmd/toxiproxy.go
@@ -3,17 +3,21 @@ package main
 import (
 	"flag"
 	"math/rand"
+	"os"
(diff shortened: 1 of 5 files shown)


proxy_collection.go
```diff
@@ -61,6 +66,49 @@ func (collection *ProxyCollection) AddOrReplace(proxy *Proxy, start bool) error
 	return nil
 }
 
+func (collection *ProxyCollection) PopulateJson(data io.Reader) ([]*Proxy, error) {
```
sirupsen: I don't think this should care about the `io.Reader`, it should probably just get the input struct directly regardless of serialization format.

cmd/toxiproxy.go
```diff
@@ -3,23 +3,49 @@ package main
...
+	flag.StringVar(&config, "config", "", "JSON file containing proxies to create on startup")
 	seed := flag.Int64("seed", time.Now().UTC().UnixNano(), "Seed for randomizing toxics with")
 	flag.Parse()
 	rand.Seed(*seed)
 }
 
 func main() {
 	server := toxiproxy.NewServer()
+	if len(config) > 0 {
+		file, err := os.Open(config)
```
sirupsen: Could we throw this in a helper method to avoid polluting `main()` too much?

proxy_collection.go
xthexder: This mimics the `AddToxicJson` and `UpdateToxicJson` functions in `toxic_collection.go` fairly well, where there's some go+json hackery going on to get things in the right format.
The `input` struct is fairly json specific since getting default vaults in a json array is a bit tricky. I'd rather not be passing it around.
At least this way it's isolated fairly well to one function.

sirupsen: Aaah... I see. Then this is fine.

sirupsen:
No objections 👍

sirupsen:
Did you ever find out why adding 100s of proxies at once is slow? I think this is still a significant amount of boot in Shopify.

xthexder:
The new `/popluate` endpoint in 2.1.0 will speed that up. It was 4x faster via the go client, and I expect it to be an even bigger difference via the ruby client.
It's able to create 1000s of proxies in less than a second, so there should be no problems there.

sirupsen:
Perfect. Can't wait to get us on this!

Issue #159, opened 2017-01-12
Toxiproxy hangs


archelan:
Scenario is:

- Add proxy
`toxiproxy-cli create proxy_20000_to_20001 --listen localhost:20000 --upstream localhost:20001`

- Add latency toxic (also reproducing with others like bandwidth)
`toxiproxy-cli toxic add proxy_20000_to_20001 --upstream --toxicName LatencyIssue -t latency -a latency=500 -a jitter=0`

- Start sending some data through, i.e.
```
while true; do; netcat -l -p 20001; done
for i in {1..100000}; do sleep 0.1; echo "$i" | tee /dev/tty ; done | netcat localhost 20000
```

- Add timeout toxic (data flow stops as expected)
`toxiproxy-cli toxic add proxy_20000_to_20001 --upstream --toxicName TimeoutIssue -t timeout -a timeout=0`

- Remove first toxic
`toxiproxy-cli toxic delete proxy_20000_to_20001 --toxicName LatencyIssue`

- As a result last command just hangs. If you try to run other commands like 'toxiproxy-cli list' they will hang as well. I'm using latest 2.1.0 release of toxiproxy and latest 1.19.1 release of cli.

sirupsen:
Ugh, I don't have time to personally debug this currently. Do you have time to dig into this? I'm considering rolling back the 2.1.0 release for the time being.

xthexder:
I have a fix for this, but I don't have push access anymore. I can create a
PR later this week when I have time.

On Feb 1, 2017 7:13 PM, "Simon Eskildsen" <notifications@github.com> wrote:

Ugh, I don't have time to personally debug this currently. Do you have time
to dig into this? I'm considering rolling back the 2.1.0 release for the
time being.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Shopify/toxiproxy/issues/159#issuecomment-276825867>,
or mute the thread

sirupsen:
Amazing @xthexder! 🎉 😍 

I've given you `admin` access to the repo. 👍

sirupsen:
@xthexder do you need anything else from me?

xthexder:
Sorry for the delay on this, I've been basically offline for the last 2 weeks due to school events and assignments.

I had a deeper look into this, and while I have a somewhat temporary fix for this, it's still not a perfect one.
The timeout toxic goes against the architectural design assumption that a toxic will always be able to write output.
[...]

PR #168, opened 2017-03-28
Timeout hang bug

jpittis:
This is a proposed fix to the API hang bug from issue https://github.com/Shopify/toxiproxy/issues/159.

### Overview of what this PR does.
- [Adds a regression test for the bug mentioned above.](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/toxics/timeout_test.go#L12)
- Fixes the bug by [changing the `timeout` toxic to drop incoming tcp data](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/toxics/timeout.go#L22-L31) on the ground.
- The `timeout` toxic implements [`CleanupToxic`](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/toxics/toxic.go#L29-L32) and [closes it's connection](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/toxics/timeout.go#L36-L38) on cleanup to avoid stream corruption because of dropped data.
- Toxiproxy calls `Cleanup` when a toxic is deleted and it implements the [`CleanupToxic`](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/toxics/toxic.go#L29-L32) interface.
- [Adds a test to confirm open connections are closed when a `timeout` toxic is deleted.](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/toxics/timeout_test.go#L63)
- [Fixes a test so that it fails gracefully rather than hanging forever.](https://github.com/Shopify/toxiproxy/blob/f7154cc4ffa799707070e94f671e3abf1fbcace8/link_test.go#L194-L211)

### TLDR of Bug (Described in issue https://github.com/Shopify/toxiproxy/issues/159.)
1. Toxics that occur before the `timeout` toxic attempt to forward TCP data to the `timeout` toxic.
2. Because the `timeout` toxic does not read incoming data, these forwarding toxics block forever.
3. When an API call deletes a toxic, it attempts to interrupt the toxic but this interrupt hangs forever because the toxic is blocked attempting to forward data to the `timeout` toxic.

### TLDR of Fix (Described in comment https://github.com/Shopify/toxiproxy/issues/159#issuecomment-287588933.)
We make the `timeout` toxic read incoming data so that writers don't block forever. The `timeout` toxic then ignores this data ("drops it on the ground"). Because the data is dropped on the ground, this corrupts the TCP stream so the connection needs to be closed when the `timeout` toxic is deleted.

### Does this break backwards compatibility?
Yes but not in realistic use cases. If you remove a `timeout` toxic, connections to this proxy will be closed. Before this change, data blocked by the `timeout` toxic would be forwarded after its deletion. (Assuming this hang bug was not encountered.). IMO the new behaviour is the correct behaviour because `timeout` implies the connection will be closed. If you want long delays, use a large value on a `latency` toxic.

### DIFF
--- a/link_test.go
+++ b/link_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"io"
 	"testing"
+	"time"
 
 	"github.com/Shopify/toxiproxy/stream"
 	"github.com/Shopify/toxiproxy/toxics"
@@ -191,13 +192,22 @@ func TestToxicity(t *testing.T) {
 	collection.chainUpdateToxic(toxic)
 
 	// Toxic should timeout after 100ms
-	n, err = link.input.Write([]byte{42})
-	if n != 1 || err != nil {
-		t.Fatalf("Write failed: %d %v", n, err)
-	}
-	n, err = link.output.Read(buf)
-	if n != 0 || err != io.EOF {
-		t.Fatalf("Read did not get EOF: %d %v", n, err)
+	done := make(chan struct{})
+	go func() {
+		n, err = link.input.Write([]byte{42})
+		if n != 1 || err != nil {
+			t.Fatalf("Write failed: %d %v", n, err)
+		}
+		n, err = link.output.Read(buf)
+		if n != 0 || err != io.EOF {
+			t.Fatalf("Read did not get EOF: %d %v", n, err)
+		}
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(150 * time.Millisecond):
+		t.Fatalf("Expected timeout after 100ms")
 	}
 }
 
--- a/link_test.go
+++ b/link_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/Shopify/toxiproxy/stream"
+	"github.com/Shopify/toxiproxy/testhelper"
 	"github.com/Shopify/toxiproxy/toxics"
 )
 
@@ -191,9 +192,7 @@ func TestToxicity(t *testing.T) {
 	toxic.Toxic.(*toxics.TimeoutToxic).Timeout = 100
 	collection.chainUpdateToxic(toxic)
 
-	// Toxic should timeout after 100ms
-	done := make(chan struct{})
-	go func() {
+	err = testhelper.TimeoutAfter(150*time.Millisecond, func() {
 		n, err = link.input.Write([]byte{42})
 		if n != 1 || err != nil {
 			t.Fatalf("Write failed: %d %v", n, err)
@@ -202,12 +201,9 @@ func TestToxicity(t *testing.T) {
 		if n != 0 || err != io.EOF {
 			t.Fatalf("Read did not get EOF: %d %v", n, err)
 		}
-		close(done)
-	}()
-	select {
-	case <-done:
-	case <-time.After(150 * time.Millisecond):
-		t.Fatalf("Expected timeout after 100ms")
+	})
+	if err != nil {
+		t.Fatal(err)
 	}
 }
 
--- a/toxics/timeout.go
+++ b/toxics/timeout.go
@@ -12,17 +12,29 @@ type TimeoutToxic struct {
 func (t *TimeoutToxic) Pipe(stub *ToxicStub) {
 	timeout := time.Duration(t.Timeout) * time.Millisecond
 	if timeout > 0 {
-		select {
-		case <-time.After(timeout):
-			stub.Close()
-			return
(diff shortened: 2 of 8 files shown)


toxics/toxic.go
```diff
@@ -26,6 +26,11 @@ type Toxic interface {
 	Pipe(*ToxicStub)
 }
 
+type CleanupToxic interface {
```
sirupsen: This is a clean approach. I like it.

toxics/timeout.go
```diff
@@ -12,19 +12,31 @@ type TimeoutToxic struct {
...
+			case <-time.After(timeout):
+				stub.Close()
+				return
+			case <-stub.Interrupt:
+				return
+			case <-stub.Input:
+				// Drop the data on the ground.
+			}
+		}
+	} else {
```
sirupsen: Why would there not be a timeout? This is essentially a "drop data" proxy in that case, right? Perhaps it should be created as a different proxy and have a different name? We could also just alias the name.

toxics/timeout_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	proxy := NewTestProxy("test", ln.Addr().String())
+	proxy.Start()
+	defer proxy.Stop()
+
+	proxy.Toxics.AddToxicJson(ToxicToJson(t, "might_block", "latency", "upstream", &toxics.LatencyToxic{Latency: 10}))
+	proxy.Toxics.AddToxicJson(ToxicToJson(t, "to_delete", "timeout", "upstream", &toxics.TimeoutToxic{Timeout: 0}))
+
+	serverConnRecv := make(chan net.Conn)
+	go func() {
```
sirupsen: It almost seems at this point we need a helper for the pattern of starting a server and accepting connections, as it's somewhat common and should be encouraged.

toxics/timeout_test.go
```diff
@@ -0,0 +1,117 @@
...
+	defer conn.Close()
+
+	_ = <-serverConnRecv
+
+	_, err = conn.Write([]byte("hello"))
+	if err != nil {
+		t.Fatal("Unable to write to proxy", err)
+	}
+
+	time.Sleep(1 * time.Second) // Shitty sync waiting for latency toxi to get data.
```
sirupsen: Why is it necessary for waiting for the latency toxic to get data to shut it down? Isn't this a race that could have problems?

toxics/timeout.go
jpittis: This is keeping consistent with the previous implementation of the `timeout` toxic. Where a `0` timeout value immediately stops all data.

I agree that the name `timeout` might be confusing. I would propose we open an issue unless you are confident in how this should be renamed / changed.

toxics/timeout_test.go
jpittis: I agree. I'm going to refactor this before shipping.

toxics/timeout_test.go
jpittis: This frustrates me.

Our bug only occurs when the `latency` toxic is blocked attempting to send data to the `timeout` toxic. Without the sleep, the test removes the `latency` toxic before it receives data which means the bug does not occur because the `latency` toxic is not blocking.

I can't think of a non racy way to wait for the `latency` toxic to receive input data. We can't wait for the upstream to receive the data because the `timeout` toxic is blocking data from reaching the upstream.

link_test.go
```diff
@@ -191,13 +192,22 @@ func TestToxicity(t *testing.T) {
...
 
 	// Toxic should timeout after 100ms
-	n, err = link.input.Write([]byte{42})
-	if n != 1 || err != nil {
-		t.Fatalf("Write failed: %d %v", n, err)
-	}
-	n, err = link.output.Read(buf)
-	if n != 0 || err != io.EOF {
-		t.Fatalf("Read did not get EOF: %d %v", n, err)
+	done := make(chan struct{})
```
jpittis: I'm also going to refactor this into a timeout test helper.

jpittis:
@sirupsen: d9e6d26 and 6b5d43d add test helpers. I decided to stick the `TimeoutAfter` helper into a generalized `testhelper` package because I find that the timeout pattern comes up all the time when testing concurrent go code.

PR #169, opened 2017-04-04
timeout toxic drops more than just the first segment

jpittis:
https://github.com/Shopify/toxiproxy/pull/168 introduced a bug where `timeout` toxics would drop the first segment it received but not future segments.

This is because I did not put a for loop around the select statement.

Quite amazingly, this was NOT caught by the regression test because the regression test only sends one segment rather than multiple.

### What does this PR do?

- Updates the regression test to send multiple segments.
- Wraps the `timeout` select statement in a for loop.

Here's a screen shot of me attempting to recreate the bug with this branch. Notice that the client does not hang when deleting the latency toxic, proving that https://github.com/Shopify/toxiproxy/issues/159 was indeed fixed.

screen shot 2017-04-04 at 7 26 21 pm

@sirupsen

### DIFF
--- a/toxics/timeout_test.go
+++ b/toxics/timeout_test.go
@@ -70,16 +70,17 @@ func WithEstablishedProxy(t *testing.T, f func(net.Conn, net.Conn, *toxiproxy.Pr
 func TestTimeoutToxicDoesNotCauseHang(t *testing.T) {
 	WithEstablishedProxy(t, func(conn, _ net.Conn, proxy *toxiproxy.Proxy) {
 		proxy.Toxics.AddToxicJson(ToxicToJson(t, "might_block", "latency", "upstream", &toxics.LatencyToxic{Latency: 10}))
-		proxy.Toxics.AddToxicJson(ToxicToJson(t, "to_delete", "timeout", "upstream", &toxics.TimeoutToxic{Timeout: 0}))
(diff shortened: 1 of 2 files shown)


toxics/timeout_test.go
```diff
@@ -70,16 +70,17 @@ func WithEstablishedProxy(t *testing.T, f func(net.Conn, net.Conn, *toxiproxy.Pr
 func TestTimeoutToxicDoesNotCauseHang(t *testing.T) {
 	WithEstablishedProxy(t, func(conn, _ net.Conn, proxy *toxiproxy.Proxy) {
 		proxy.Toxics.AddToxicJson(ToxicToJson(t, "might_block", "latency", "upstream", &toxics.LatencyToxic{Latency: 10}))
-		proxy.Toxics.AddToxicJson(ToxicToJson(t, "to_delete", "timeout", "upstream", &toxics.TimeoutToxic{Timeout: 0}))
+		proxy.Toxics.AddToxicJson(ToxicToJson(t, "timeout", "timeout", "upstream", &toxics.TimeoutToxic{Timeout: 0}))
```
jpittis: I renamed this because the name `to_delete` had no meaning in this context as the `timeout` toxic is not being deleted in this test.

PR #181, opened 2017-05-25
Handle SIGTERM to exit cleanly

### LINKED ISSUE #180: toxiproxy Server should exit with code 0 when SIGTERM'd
Currently, when toxiproxy Server receives a SIGTERM signal (from `docker stop` say) it exits with a non-zero code. 

Is this desired, or a feature? Could this be changed by handling a SIGTERM and exiting with 0?

dlmiddlecote:
Fixes #180

### DIFF
--- a/cmd/toxiproxy.go
+++ b/cmd/toxiproxy.go
@@ -3,6 +3,9 @@ package main
 import (
 	"flag"
 	"math/rand"
+	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jpittis:
That's the "right thing to do" but I can't imagine the use case. Toxiproxy only ever has a single user. (The individual or bot who is running the test suite.) If that user wants to kill Toxiproxy, why would they ever care about leftover state in the pipes?

sirupsen:
Agree 👍  Let's ship it

PR #184, opened 2017-06-22
Address an issue where a toxiproxy can be used to bypass the Same-Origin Policy in web browsers

JackMc:
This PR provides a temporary fix to an issue where proxies can be created from the browser.

### DIFF
--- a/api.go
+++ b/api.go
@@ -7,10 +7,11 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/Shopify/toxiproxy/toxics"
-	"github.com/sirupsen/logrus"
 	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
 )
 
 type ApiServer struct {
@@ -46,6 +47,16 @@ func (server *ApiServer) PopulateConfig(filename string) {
 	}
 }
 
+func StopBrowsersMiddleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.HasPrefix(r.UserAgent(), "Mozilla/") {
+			http.Error(w, "User agent not allowed", 403)
+		} else {
+			h.ServeHTTP(w, r)
+		}
+	})
+}
+
 func (server *ApiServer) Listen(host string, port string) {
 	r := mux.NewRouter()
 	r.HandleFunc("/reset", server.ResetState).Methods("POST")
@@ -62,7 +73,8 @@ func (server *ApiServer) Listen(host string, port string) {
 	r.HandleFunc("/proxies/{proxy}/toxics/{toxic}", server.ToxicDelete).Methods("DELETE")
 
 	r.HandleFunc("/version", server.Version).Methods("GET")
-	http.Handle("/", r)
+
+	http.Handle("/", StopBrowsersMiddleware(r))
 
 	logrus.WithFields(logrus.Fields{
 		"host":    host,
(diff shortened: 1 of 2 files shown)


JackMc:
Reference for most user agents starting with Mozilla: 
screen shot 2017-06-22 at 2 15 30 pm

(Note that the Presto engine is not used in modern browsers)

api.go
```diff
@@ -46,6 +47,16 @@ func (server *ApiServer) PopulateConfig(filename string) {
 	}
 }
 
+func StopBrowsersMiddleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.HasPrefix(r.UserAgent(), "Mozilla/") {
```
jpittis: Is matching against browser user agents that simple?

JackMc: According to MDN, all the common browsers' user agents start with `Mozilla/`

JackMc: http://webaim.org/blog/user-agent-string-history/ User agents have a sad history of pretending to be each other

EiNSTeiN-:
> Is this the standard way to solve such issues? This seems kind of.. awkward? No concerns otherwise.

The standard way to solve this issue would be to require clients to provide a custom HTTP header. This can't be done by browsers  (it requires CORS, which wouldn't be allowed in this case) so it would prove that the request comes from a client that isn't a browser.

xthexder:
I'm curious about what particular scenario this is trying to prevent?
Access from the broswer is nice to have, and by design in the case of this dashboard PR https://github.com/Shopify/toxiproxy/pull/218

You can only create and connect to proxies from the same machine running toxiproxy, which means there would have to be malicious code running on localhost (or I guess the test CI server). Toxiproxy shouldn't be running in production environments, and even if it was, it definitely shouldn't be publicly accessible.
Even if you could create a proxy to access another site, you wouldn't have access to the browser's session still.

Maybe there's some scenario I'm missing? If this is still important we can add some CSRF token stuff to the dashboard PR.

EiNSTeiN-:
Same-Origin policy is one of the cornerstone of browser security. Browsers enforce that a website example.com can only read data from itself, not from google.com. For obvious reasons bypassing the same-origin policy would be a major problem.

Toxiproxy doesn't know which domain name it is served from, so IIRC this PR aimed at preventing browsers from connecting to it. Without this, a dev running a local toxiproxy visiting a random evil.com site could get its proxy configuration modified by evil.com to access any local resources that the developer has access to on the local network.

JackMc:
Hi @xthexder! Thanks for the comment.

As @EiNSTeiN- said, this was introduced to stop violations of the same origin policy on developer machines through abuse of Toxiproxy.

You're correct that this could be fixed by introducing a CSRF token, but the problem there is that it would require modifications to several third party clients and a breaking change. Therefore, we decided to go with this approach at least in the short term.

xthexder:
In that example, how does `evil.com` access `localhost:8474`? The very same same-origin policy would block it. Either toxiproxy has to be accessible through `evil.com`, or the malicious code has to be running on localhost. There's also nothing stopping `evil.com` from hosting its own proxy via nginx or any other method. I'm still failing to come up with a particular scenario that is a problem.

EiNSTeiN-:
> The very same same-origin policy would block it.

[DNS Rebinding](https://www.tripwire.com/state-of-security/vert/practical-attacks-dns-rebinding/) can be used when the server does not validate the domain name used to connect to it (which is the case for toxiproxy AFAIK)

xthexder:
Ah yeah, that's the loophole I was looking for. I guess browser access is staying off until an auth token of some sort is added.

Maybe in the next major version rev we can fix this properly with updates to the clients.

PR #191, opened 2017-10-10
only prepend `http://` to endpoint if no protocol was provided

### LINKED ISSUE #174: cli fails with https hosts
```
toxiproxy-cli --host https://my-toxyproxy-host.com create toxi_name -l localhost:8000 -u my-upstream-host
Failed to create proxy: Post http://https//my-toxyproxy-host.com/proxies: dial tcp: lookup https: no such host
```

For some reason, the `--host` value is appended with `http://` (`http://https//my-toxyproxy-host.com`). It works fine if it already starts with `http://` though.

Crunch09:
Hey,
this addresses the fix mentioned in #174: Although the Toxiproxy HTTP API does not support HTTPS, it should not prepend `http://` to the endpoint.

```go
client.NewClient("https://example.com")
// Before: http://https//example.com
// After: https://example.com
```
I would love to write a test for this but `endpoint` is not an exported identifier so i'm not sure how to test this as this is basically the first Go code that i have ever written 🙂 

fixes #174

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -11,7 +11,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
-	"strings"
+	"regexp"
 )
 
 // Client holds information about where to connect to Toxiproxy.
@@ -47,7 +47,8 @@ type Proxy struct {
 // with Toxiproxy. Endpoint is the address to the proxy (e.g. localhost:8474 if
 // not overriden)
 func NewClient(endpoint string) *Client {
-	if !strings.HasPrefix(endpoint, "http://") {
+	protocolProvided, _ := regexp.MatchString(`^https?://`, endpoint)
+	if !protocolProvided {
 		endpoint = "http://" + endpoint
 	}
 	return &Client{endpoint: endpoint}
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -47,7 +47,8 @@ type Proxy struct {
 // with Toxiproxy. Endpoint is the address to the proxy (e.g. localhost:8474 if
 // not overriden)
 func NewClient(endpoint string) *Client {
-	if !strings.HasPrefix(endpoint, "http://") {
+	protocolProvided, _ := regexp.MatchString(`^https?://`, endpoint)
```
jpittis: Generally it's better to create a regex via `regexp.MustCompile()` and then calling the `MatchString` method. This ensures we don't ignore a possible error in the regex.

Crunch09:
@jpittis Thank you for the review! 👍  I updated the code to now fail when `https://` is provided and removed the regexp again because i felt it didn't add much value then but thank you for letting me know about `MustCompile`! 💯  

**Before**:
bildschirmfoto 2017-10-12 um 23 17 07

**After**:
bildschirmfoto 2017-10-12 um 23 17 14

jpittis:
Thanks for the contribution! :)

PR #194, opened 2017-10-23
Dashboard

### LINKED ISSUE #192: Dashboard
Following a [twitter chat with @sirupsen](https://twitter.com/jacobbednarz/status/917572306787954688), I'm going to start looking at adding a UI component to core. 

The general look will be something like the Shopify interface since this seems to cover off all the points that I can see being needed.

My needs are going to be quite basic at this stage but is there any considerations you'd like to add before I embark too far on this journey?

jacobbednarz:
This is the start of a UI for Toxiproxy ripped heavily from the existing Shopify app but built into Toxiproxy core instead of an add-on.

### Checklist

- [x] Have all Proxies + Toxics shown
- [ ] Be able to update the Proxy or Toxic using the UI without touching the CLI
- [ ] Pretty up the design

This is still verrrry WIP however I thought I would PR this early in case I either don't end up finishing it or anyone has any input during the journey. 

Fixes #192

### DIFF
--- a/api.go
+++ b/api.go
@@ -74,6 +74,8 @@ func (server *ApiServer) Listen(host string, port string) {
 
 	r.HandleFunc("/version", server.Version).Methods("GET")
 
+	r.HandleFunc("/dashboard", server.Dashboard).Methods("GET")
+
 	http.Handle("/", StopBrowsersMiddleware(r))
 
 	logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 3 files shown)


api.go
```diff
@@ -389,6 +392,16 @@ func (server *ApiServer) Version(response http.ResponseWriter, request *http.Req
 	}
 }
 
+func (server *ApiServer) Dashboard(response http.ResponseWriter, request *http.Request) {
+	var dashboardData []proxyToxics
```
jacobbednarz: I made a decision early on to not use the existing HTTP API for these since we already have a couple of functions  handy for fetching them.

templates/dashboard.tmpl
```diff
@@ -0,0 +1,55 @@
...
+            toxicity: {{ .Toxicity }}
+        </td><td>
+            direction: {{ .Direction }}
+        </td><td>
+            index: {{ .Index }}
+        </td><td>
+            buffer: {{ .BufferSize }}
+        </td><td>
+        </td><td>
+            {{ if eq .Type "latency" }}
```
jacobbednarz: these `if` blocks will probably go away once the updating work comes into it.

jacobbednarz:
Closing this as #218 is further along

PR #195, opened 2017-11-16
Introduce a toxic that closes a connection after given time period

mgkeen:
Whilst using toxiproxy to do some testing, i found the timeout toxic didn't quite suit my needs. 
If i understand it correctly, the timeout toxic doesn't allow data through at all. I needed a toxic that would allow a connection to carry data, and then cut off after some amount of time.
I took inspiration from the limit_data toxic and introduced a limit_time toxic. There's probably a better name for it that i haven't thought of.

### DIFF
--- a/toxics/limit_time.go
+++ b/toxics/limit_time.go
@@ -1,0 +1,48 @@
+package toxics
+
+import "time"
+
+// LimitTimeToxic has shuts connection after given time
+type LimitTimeToxic struct {
+	Time int64 `json:"time"`
+}
+
+type LimitTimeToxicState struct {
+	ElapsedMilliseconds int64
+}
+
+func (t *LimitTimeToxic) Pipe(stub *ToxicStub) {
+	state := stub.State.(*LimitTimeToxicState)
+
+	if state.ElapsedMilliseconds >= t.Time {
+		stub.Close()
+		return
+	}
+
+	timeout := time.Duration(t.Time-state.ElapsedMilliseconds) * time.Millisecond
+	start := time.Now()
+	for {
+		select {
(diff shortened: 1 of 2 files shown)


jpittis:
Thanks for the PR @mgkeen. I'll try to look at it at some point next week.

customcoops77:
Yep I know that I don't know how to do this. Been trying to get a pull forever. Can I get help or do I suffer like it has. Please need help.

sl1pm4t:
Thanks for the useful addition @mgkeen.
I found one issue with our use case.
The current implementation doesn't work as expected (connection is not killed on timeout) if there is constant traffic on the connection (as was happening in my case with a WebSocket connection that is sending ping / pong every second).

The stub select case: `case c := <-stub.Input:` trips before any other and elapsed time is not recalculated.

xmo-odoo:
Any news on this? I've got to say this is the behaviour I was expecting "timeout" to have, and the cli help didn't make it clear it would prevent all data passthrough entirely.

Though maybe this could be merged into timeout and the traffic block behaviour would be either a separate toxic to combine with, or an attribute on the toxic? In fact it seems to me traffic blocking could be done by setting `bandwidth` to 0.

Though I guess if you're using the chaotic mode (toxicity != 1) you may want the toxics to always be enabled together… maybe the toxics should be more orthogonal but the toxicity should apply to groups of toxics rather than individual ones?

mgkeen:
As the previous comments suggest my implementation here probably doesn't work all that well. I might be able to have a go at doing this properly at some point soon, but I'm quite busy these days so there's a good chance that won't happen. If anybody else wants to pick it up then feel free!

PR #204, opened 2018-03-04
add support for toxiproxy-haskell

jpittis:
I've written a Haskell Toxiproxy client.

It's well tested. (Complete test suite for every endpoint of the API.) So I believe it's appropriate to add to the list of client libraries.

The Haskell HTTP client library I used requires the "text/plain" content type to be labeled with a charset. I think it's reasonable that we expect our version number to be valid utf-8. I've updated the `Content-Type` field of `Version` to specify a charset.

@sirupsen


jpittis:
❤️ Thanks Jacob!

sirupsen:
I'm so ambivalent, but you know I love you jake ❤️

Issue #79, opened 2015-08-28
Toxics are not set when creating a proxy for the first time.


jpittis:
When using the Go client as follows, toxics are not added to a proxy.
Given `var p []toxiproxy.Proxy` with toxics.
```` Go
proxy := client.NewProxy(&p)
proxy.Create()
````
This does not properly add the configured toxics from `p` onto the proxy.

They must be added after the fact using `Proxy.SetToxic`.

Sirupsen:
yeah this is a known "issue". @xthexder I don't think this is any different in 2.0?

xthexder:
Yeah, it's not any different currently, just a little less misleading. I think it would be reasonable to allow toxics to be specified in the `Populate()` function

PR #211, opened 2018-04-15
Populate toxics at startup from a json string

worldtiki:
# Ability to provide a configuration at startup.
My use case is the following:
I'm running a Kuberneters cluster with several applications/containers deployed. I would like to use Toxiproxy as a sidecar to some of these applications to be able to perform failure testing.

Currently we have two existing options to populate proxies, one from a json file and by making an http request. Neither option is very friendly when using Toxiproxy on containerized environments as they either require a post start script to run the http populate request or mounting a volume with the json config file.

With this change we can provide a json string (at startup) as such:
`docker run -it shopify/toxiproxy:git -configJson '[{"name": "redis","listen": "0.0.0.0:9092","upstream": "redis:6379"}]'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,8 +288,11 @@ Toxiproxy.
 For large application we recommend storing the Toxiproxy configurations in a
 separate configuration file. We use `config/toxiproxy.json`. This file can be
 passed to the server using the `-configFile` option, or loaded by the application
-to use with the `populate` function. Alternatively you can pass a json string 
-with the `-configJson` option to populate these configurations.
+to use with the `populate` function.
+Alternatively you can pass a json string with the `-configJson` option to populate
+these configurations. This is specially useful when running Toxiproxy as a docker container, e.g.:
+
+```docker run -it shopify/toxiproxy -configJson '[{"name": "redis","listen": "0.0.0.0:9092","upstream": "redis:6379"}]'```
 
 Use ports outside the ephemeral port range to avoid random port conflicts.
 It's `32,768` to `61,000` on Linux by default, see
--- a/README.md
+++ b/README.md
@@ -287,8 +287,9 @@ Toxiproxy.
 
 For large application we recommend storing the Toxiproxy configurations in a
 separate configuration file. We use `config/toxiproxy.json`. This file can be
-passed to the server using the `-config` option, or loaded by the application
-to use with the `populate` function.
+passed to the server using the `-configFile` option, or loaded by the application
+to use with the `populate` function. Alternatively you can pass a json string 
+with the `-configJson` option to populate these configurations.
 
 Use ports outside the ephemeral port range to avoid random port conflicts.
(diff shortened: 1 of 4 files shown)


jpittis:
Hey @worldtiki. Are you attempting to run Toxiproxy as a side car in a production Kubernetes cluster? If so, I would suggest extreme caution. Toxiproxy was not designed for production use and I have never heard of anyone running it in production environments.

Thanks for opening a PR. I don't have time to review this right now but will look at it later in the week.

worldtiki:
Hi @jpittis. Thanks for the feedback. This would be used in a staging/test environment, not in a production one.

jpittis:
Did you find an alternative solution @worldtiki? If you did, would you mind briefly mentioning what it was?

jpittis:
The correct behaviour is that initial population ignores toxics. There's an open issue. https://github.com/Shopify/toxiproxy/issues/79

jpambrun:
@worldtiki, Any chance to revive that? I like what you have done on your fork.

jpambrun:
However, I just noticed that the ability to update toxics through the api seems to be broken on your branch.

worldtiki:
Yeah, sorry about that @jpambrun :(
I needed some extra features so I kinda butchered the project in that branch.

I'll reopen this pr (didn't know it could be of use for others). Regarding the initial population of toxics, it would be nice if someone else could give it a go, as my knowledge of go is very basic :s

marqub:
@jpambrun @worldtiki 
If you  still want to use ToxiProxy with k8s, I come up with a solution that does not require code changes in ToxiProxy
Here a story on Medium that details the approach
http://bit.ly/2EAXRYk

Hope it helps!

miry:
I would replace ability to load config via ENV variable instead. Currently I am not fully convinced to change the arguments.

miry:
@worldtiki I tried to find other command lines that uses ENV as full config or for complex one. I failed to find a good examples (kubectl, opentelemtry-collector, grafana, nginx).  Learned about best practices in https://12factor.net/ and https://clig.dev/.

Max that we can do assign SINGLE ENV to SINGLE Flag. Similar to TOML or what provides current frameworks [Viper](https://github.com/spf13/viper) or https://github.com/urfave/cli/.

As workaround for docker we can use entrypoint.sh that would generate config file base on ENVs.

Example: https://devopsian.net/notes/docker-nginx-template-env-vars/  or use http://mustache.github.io/mustache.5.html

P.S: During learning about config structure. I feel I need extract parse config out of models. I am going to change config structure to remove arrays of proxies with dictionary. So it would be easy to create new entries. Also support more other formats.

miry:
Also thought to use flag `--proxy` and to specify attributes with `"upastream=local:2,downstream=local,name=oops,enabled=true"` 

Also i think it is not unix way.

PR #215, opened 2018-05-31
(upstream) toxic for http headers

worldtiki:
Raising this pr and hoping to get some traction on Toxiproxy again :)

**Context**: it's common to use the host header to direct a http request to a particular service, specially when using load balancers/reverse proxies. This is a problem with toxiproxy as the victim app will be configured to send the requests to localhost:port and when toxiproxy forwards it to the reverse proxy the host header will be localhost and the request will fail to be delivered to the correct service.

This toxic is an upstream toxic only and allows forcing http headers for a proxy (host and others). I understand there's been some discussion about bidirectional toxics but that work was not done yet, and I think it would be of interest to solve this issue now and refactor this solution later to add downstream support


xthexder:
Actually, I think most of the work on bidirectional toxics was done, it just needed some review and polishing. I got stalled and stopped working on it, but I could put the time in if there's interest.

I'd love for toxiproxy to start supporting protocol aware toxics, since I think it's a powerful tool, and would have a lot of applications.

nikos912000:
Hey Toxiproxy team,
Is there any update on this? We're using Toxiproxy a lot but without forwarding the host header our requests fail to get forwarder to the right service.

jpittis:
Regarding bidirection toxics, as @xthexder explained the work is mostly done. Someone needs to own understanding + polishing + shipping it. Currently no one is owning that work though. (https://github.com/Shopify/toxiproxy/pull/132)

Regarding this PR, I don't see the issue with adding a "hackier" short term solution in Toxic form. Especially since toxics are well de-coupled from the rest of the system. We should get this PR reviewed.

toxics/http_request_headers_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for k, v := range r.Header {
+		// headers can contain multiple elements. for the purposes of this test we pick the 1st
+		headersMap[k] = v[0]
+	}
+
+	mapAsJson, _ := json.Marshal(headersMap)
+	w.Write([]byte(mapAsJson))
+}
+
+func TestToxicAddsHttpHeaders(t *testing.T) {
```
worldtiki, line 27:
```suggestion
func TestToxicAddsHTTPHeaders(t *testing.T) {
```

worldtiki:
Hi folks! Is there still interest in this?

avaakash:
Hi @worldtiki, I was trying out this toxic, but not able to pass headers as an attribute, there is a problem in unmarshalling. I wanted to know how were you creating this toxic using the CLI.
Command I used in CLI
` ./toxiproxy-cli create -l localhost:2002 -u wttr.in:80 test && ./toxiproxy-cli toxic add -t response_header -a headers={"new":"test"} test`

Error I am getting
`Failed to add toxic: failed to add toxic to proxy test: AddToxic: HTTP 400: bad request body: json: cannot unmarshal string into Go struct field HeaderToxic.attributes.headers of type map[string]string`

worldtiki:
@dianadevasia ?

dianadevasia:
@worldtiki : Sorry the PR got accidentally closed when the branch got renamed. I have reopened the PR and updated the source.

PR #220, opened 2018-07-22
Update blog url

pulkitsharma07:


jpittis:
One redirects to the other so I'm not sure this is a big deal.

But I'm 👍with this.

Issue #244, opened 2019-02-11
Upstream aware toxics


chbatey:
I'd like to be able to write toxics that are upstream aware e.g. drop if client's IP = x.

This will be useful for testing peer 2 peer systems rather than client server. 

The usecase I have in mind is for a service that forms a cluster e.g. a Cassandra Cluster or an Akka Cluster I'd like to have toxi proxy between each P2P connection and be able to simulate a node misbehaving. As these are P2P systems each node will connect to a subset of other nodes so I only wan to disrupt incoming traffic from a given source.

To support this I think we'd need to add more information to `ToxicStub` or create a new toxic type. 

WDYT? Is this something you'd accept if I worked on it?

jpittis:
Heya @chbatey!

One way to do that would be to add more information about the connection to the [ToxicStub](https://github.com/Shopify/toxiproxy/blob/7eb6f30c2c5a2b750732be86b7717b1d6970c879/toxics/toxic.go#L58-L65) as you suggest. I can't think of a better approach. I think we'd accept something like this.

(Mostly unrelated, FYI that https://github.com/Shopify/toxiproxy/issues/50 protocol aware toxics have been a WIP for a while.)

chbatey:
Thanks @jpittis I raised https://github.com/Shopify/toxiproxy/pull/246 and have been testing with it and some custom toxics and it works. We could instead not have a toxic added to the link rather than all the toxics that want to be client specific doing the check.

xthexder:
I'm not sure how useful this is, since it's a fairly old PR, but some of the code in https://github.com/Shopify/toxiproxy/pull/132 might be helpful for this.

PR #246, opened 2019-02-13
Add client address to ToxicStub to enable client aware toxics

chbatey:
This seemed the simplest way of doing this and I've written toxics that enable me to mimic mis-behaving nodes in an Akka Cluster (P2P clustering). 

I haven't written golang in years so apologies for any atrocities :) 

Refs #244


chbatey:
I've signed the CLA

PR #253, opened 2019-06-10
Move to go modules

paranoidaditya:
## What

Move to Go Modules from godeps


thegedge:
Everything looks 👍, but I don't do a lot of work here so let's find someone who does and see if they're happy moving to modules.

PR #257, opened 2019-07-19
Formatted code using goimports

muesli:
Code formatting / import order fixes only.

### DIFF
--- a/testhelper/testhelper.go
+++ b/testhelper/testhelper.go
@@ -1,8 +1,8 @@
 package testhelper
 
 import (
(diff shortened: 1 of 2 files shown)


muesli:
fwiw, I signed the CLA, but it seems the bot isn't updating.

miry:
@muesli thank you for your PR. It was simple and valuable proposal. I started cleaned up the project. After review  found that the files it seems formatted :(. I refreshed the status of CLA in Github checks - it is green now.

PR #264, opened 2019-11-07
Pull base image by digest and not tag

worldtiki:
Very small change. Pulling image by digest instead of tag to have a deterministic (and safer) build.

I'm using the latest alpine (https://hub.docker.com/layers/alpine/library/alpine/3.10.3/images/sha256-e4355b66995c96b4b468159fc5c7e3540fcef961189ca13fee877798649f531a)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM alpine
+# Alpine 3.10.3
+FROM alpine@sha256:e4355b66995c96b4b468159fc5c7e3540fcef961189ca13fee877798649f531a
 
 COPY tmp/build/toxiproxy-server-linux-amd64 /go/bin/toxiproxy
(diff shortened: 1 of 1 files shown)


sirupsen:
What about security updates etc?

worldtiki:
> What about security updates etc?

Good point.

The digest may be overkill but I think using a tag would be beneficial in order to have deterministic builds.

Have you considered using Renovate or Dependabot in this project? It's a way to automate dependency updates. Both support several languages/... including Docker. (example here for a Java/Maven update: https://github.com/ExpediaGroup/pitchfork/pull/196/files)

miry:
Switched to `scratch` https://github.com/Shopify/toxiproxy/blob/26f2a85bd75698427dac087860068c973ff647fa/Dockerfile#L1

PR #288, opened 2020-11-19
Adding CLI rst command

peter-ginchev:


peter-ginchev:
@jpittis Could you please review this chunk of code, and submit, if relevant, I just wanted to add RST functionality, which was useful for me.

miry:
Hi @peter-ginchev 

Thank you for your contribution. I wonder if you think your changes are still relevant, after so long time?

peter-ginchev:
> Hi @peter-ginchev
> 
> Thank you for your contribution. I wonder if you think your changes are still relevant, after so long time?

Yes, At that moment we wanted to simulate RST and we did it successfully using this patch.
I see that currently, the code has changed, and the conflict is around the fix I need to change the created client fro "net.Listen" to "net.ListenTCP", because RST is thrown on a TCP connection, and the connection previously was just IP.

miry:
Hi @peter-ginchev 

I decided to use #247 as main solution to add Reset peer toxic.
It adds toxic with timeout to create a disturbance. 

I released a new version https://github.com/Shopify/toxiproxy/releases/tag/v2.2.0 
Let me know if it solves your use case as well.

PR #290, opened 2021-01-27
Mention MacPorts in readme

amake:
toxiproxy is also available in MacPorts.


amake:
I've now signed the CLA.

miry:
Thank you very much. Sorry for such long wait.

PR #296, opened 2021-06-07
Add a systemd unit file

### LINKED ISSUE #205: Unit toxiproxy.service not found
Hi :wave: 

I'm trying to install toxiproxy following the ubuntu instructions: 
```
$ wget -O toxiproxy-2.1.3.deb https://github.com/Shopify/toxiproxy/releases/download/v2.1.3/toxiproxy_2.1.3_amd64.deb
$ sudo dpkg -i toxiproxy-2.1.3.deb
$ sudo service toxiproxy start
```

But i'm getting `Failed to start toxiproxy.service: Unit toxiproxy.service not found.` after running `$ sudo service toxiproxy start`

Ladicek:
Co-authored-by: Jim Kleine <jimkleine@users.noreply.github.com>

Fixes #205

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,8 @@ $(DEB): tmp/build/$(SERVER_NAME)-linux-amd64 tmp/build/$(CLI_NAME)-linux-amd64
 		--url "https://github.com/Shopify/toxiproxy" \
 		$(word 1,$^)=/usr/bin/$(SERVER_NAME) \
 		$(word 2,$^)=/usr/bin/$(CLI_NAME) \
-		./share/toxiproxy.conf=/etc/init/toxiproxy.conf
+		./share/toxiproxy.conf=/etc/init/toxiproxy.conf \
+		./share/toxiproxy.service=/lib/systemd/system/toxiproxy.service
 
 docker:
 	docker build --tag="shopify/toxiproxy:git" .
--- a/share/toxiproxy.service
+++ b/share/toxiproxy.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=TCP proxy to simulate network and system conditions
+After=network-online.target firewalld.service
+Wants=network-online.target
+
+[Service]
+Type=simple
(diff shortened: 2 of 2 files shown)


jimkleine:
Hi @miry 

If you can clarify your intent, I can take a look at this (I provided the original sample systemd service file), as there has just been another report of the service failing with an error on Ubuntu 24.04.

I think you are:

1.  Proposing a PR to install the systemd service via .goreleaser.yml; and

2.  Asking that the service be disabled by default.  I might question if this is a reasonable user expectation? If yes, this implies that the documentation may also need to be updated.

Have I understood you correctly?

I don't often work with Go, so I will need a couple of days to get up to speed with Goreleaser.

> Hi @Ladicek 
> 
> thank you for the contribution. 
> I changed a bit a releasing part. 
> 
> Now there is [Goreleaser](https://goreleaser.com).
> The configuration for linux packages specified in https://goreleaser.com/customization/nfpm/
> 
> Here is an example: https://github.com/Shopify/toxiproxy/blob/master/.goreleaser.yml#L63
> 
> Is it possible to disable running service by default?

miry:
> Asking that the service be disabled by default. I might question if this is a reasonable user expectation? If yes, this implies that the documentation may also need to be updated.

As far as I know, most packages I have used on Fedora require users to configure the desired behavior themselves.

jimkleine:
@miry Thanks for your feedback. Your reference to Fedora led me to [this guidance on default service starts](https://berrange.fedorapeople.org/mingw-integrated/packaging-guidelines/DefaultServices/).   Toxiproxy might meet test #4 if it *only* listens on localhost by default.

I had been thinking about this through a developer lens, but there is a wider security implication to consider.

> > Asking that the service be disabled by default. I might question if this is a reasonable user expectation? If yes, this implies that the documentation may also need to be updated.
> 
> As far as I know, most packages I have used on Fedora require users to configure the desired behavior themselves.
>

PR #302, opened 2021-08-27
Automate release with Github actions

miry:
Use Github Container registry instead of DockerHub.


thepwagner (approved):
LGTM, just a nit that's safe to ignore.

PR #309, opened 2021-09-07
E2E tests

miry:
Add script to run e2e tests to test client and server.
Use benchmarking script as sample client application.
Automate the process with Github Actions in test workflow.


thepwagner (approved):
LGTM!

I wondered about using `docker-compose` to manage these instead:
* `wait_for_url` looks a lot like a docker `HEALTHCHECK`
* `trap pkill ; pkill` is a little like `docker-compose down -f`
* the pattern of prefixing log output from the `[web]` and `[toxiproxy`] servers with sed is also ~free in docker-compose

I'm not sold it's worth adding the extra dependency, just noting the similarities.

miry:
@thepwagner Thank you for your review and feedback.

You are right about `docker-compose`, and the reasons why I try avoid adding new tools.

PR #311, opened 2021-09-13
chore: add /v2 suffix to module import path

### LINKED ISSUE #310: toxiproxy v2.1.5 can't be used as a client module without /v2 suffix
Since https://github.com/Shopify/toxiproxy/pull/253 added a `go.mod` for toxiproxy, the v2.x.x release from v2.1.5 can no longer be referred to as an `+incompatible` module dependency and is required to have a `/v2` suffix on the module path.

```
server response: not found: github.com/Shopify/toxiproxy@v2.1.5+incompatible: 
  invalid version: +incompatible suffix not allowed: 
    module contains a go.mod file, so semantic import versioning is required
```
(see [Go Modules have a v2+ Problem](https://donatstudios.com/Go-v2-Modules) for further discussion)

dnwe:
Since https://github.com/Shopify/toxiproxy/pull/253 added a `go.mod` for
toxiproxy, the v2.x.x release from v2.1.5 can no longer be referred to
as an `+incompatible` module dependency and is required to have a `/v2`
suffix on the module path.

```
server response: not found: github.com/Shopify/toxiproxy@v2.1.5+incompatible:
  invalid version: +incompatible suffix not allowed:
    module contains a go.mod file, so semantic import versioning is required
```
(see [Go Modules have a v2+ Problem](https://donatstudios.com/Go-v2-Modules) for further discussion)

Fixes #310

### DIFF
--- a/api.go
+++ b/api.go
@@ -9,7 +9,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/Shopify/toxiproxy/toxics"
+	"github.com/Shopify/toxiproxy/v2/toxics"
 	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
 )
--- a/api_test.go
+++ b/api_test.go
@@ -7,8 +7,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/Shopify/toxiproxy"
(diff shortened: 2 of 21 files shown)


miry:
Thank you @dnwe

PR #316, opened 2021-09-16
Set max line 100 characters

miry:
Add linter to check the max line length.

Use https://github.com/segmentio/golines to format most of files.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,4 +6,5 @@ linters:
 
 linters-settings:
   lll:
-    line-length: 200
+    line-length: 100
+    tab-width: 1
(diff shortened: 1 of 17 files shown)


.golangci.yml
```diff
@@ -6,4 +6,5 @@ linters:
 
 linters-settings:
   lll:
-    line-length: 200
+    line-length: 100
+    tab-width: 1
```
lilien1010, line 10: how about adding `statements:75` too

lilien1010 (approved):
LGTM

.golangci.yml
miry: I could not find this settings in https://github.com/golangci/golangci-lint/blob/2ea496f22b997490dd43bbb1536f28908b4389c6/pkg/config/linters_settings.go#L337

miry: I think it is from number of lines for FunLength: https://github.com/golangci/golangci-lint/blob/2ea496f22b997490dd43bbb1536f28908b4389c6/pkg/config/linters_settings.go#L200

PR #323, opened 2021-09-17
Extract initialization of upstream server in tests

miry:
Extract common part in test to initialize the test upstream server.


lilien1010 (approved):
LGTM

PR #351, opened 2022-01-05
Add fix for tests on MacOS 12 with -race flag for go17

Strech:
> On macOS 12 a new malloc implementation (nano) is used by default, and apparently it reserves address range 0x600000000000-0x600020000000, which conflicts with the address range that TSAN uses for Go
> https://github.com/golang/go/issues/49138

The fix is already merged in upstream, but yet not released. Change will perform OS and Golang version check and add `MallocNanoZone=0` on MacOS with go17 <= 1.17.5 (I hope that 1.17.6 will include upstream fix)

**Question:** Do we want to include other Go versions? Or we consider 1.17 as the main one on MacOS?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # [Unreleased]
 
 * Verify git tag on release (#347, @miry)
+* Fix MacOS 12 tests for go17 with -race flag (#351 @strech)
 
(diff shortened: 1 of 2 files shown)


miry (approved):
Need to update Changelog also - and ❤️

PR #356, opened 2022-01-12
[RFC] Don't replace proxy with random port in AddOrReplace

tinnet:
When asking ProxyCollection to `AddOrReplace` a proxy with a random
listening port (`:0`), don't stop and replace an existing one with
the same `name` and `upstream`.

### DIFF
--- a/proxy_collection_test.go
+++ b/proxy_collection_test.go
@@ -41,6 +41,30 @@ func TestAddTwoProxiesToCollection(t *testing.T) {
 	}
 }
 
+func TestAddOrReplaceProxyToCollectionWithRandomPort(t *testing.T) {
+	collection := toxiproxy.NewProxyCollection()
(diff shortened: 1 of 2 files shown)


proxy_collection_test.go
```diff
@@ -41,6 +41,30 @@ func TestAddTwoProxiesToCollection(t *testing.T) {
 	}
 }
 
+func TestAddOrReplaceProxyToCollectionWithRandomPort(t *testing.T) {
+	collection := toxiproxy.NewProxyCollection()
+	proxy1 := NewTestProxy("test", "localhost:20000")
+	proxy2 := NewTestProxy("test", "localhost:0")
```
miry, line 47: As I understand the Listen for NewTestProxy is always "localhost:0". The second argument is upstream.

miry:
Clossing for now. Pls reopen if it is still required.

PR #358, opened 2022-01-19
Improve server log startup message

### LINKED ISSUE #228: Improve server log message so people know Toxiproxy is working.
https://github.com/Shopify/toxiproxy/issues/226

Maybe even: "curl http://localhost: 8474/version to ensure Toxiproxy is working." or somethiung

areveny:
Old output
```
dist % ./toxiproxy-server
INFO[0000] API HTTP server starting                      host=localhost port=8474 version=git
```

New output
```
dist % ./toxiproxy-server
INFO[0000] Toxiproxy HTTP server is starting             host=localhost port=8474 version=git
INFO[0000] Toxiproxy has started. Verify with `curl localhost:8474/version`
```

Closes #228

### DIFF
--- a/api.go
+++ b/api.go
@@ -81,7 +81,8 @@ func (server *ApiServer) Listen(host string, port string) {
 		"host":    host,
 		"port":    port,
 		"version": Version,
-	}).Info("API HTTP server starting")
+	}).Info("Toxiproxy HTTP server is starting")
+	logrus.Printf("Toxiproxy has started. Verify with `curl %s:%s/version`", host, port)
 
 	err := http.ListenAndServe(net.JoinHostPort(host, port), nil)
 	if err != nil {


miry:
Hi @areveny 

Thank you for contribution. I have not seen in other application explanation how to use health endpoint. I would avoid write this in the logs.

The application should crash if there is some problems. Also I think the second message does not correctly says that server hast started listen for the port. But it was executed just write after the first message. 

I would keep only one message:

```
INFO[0000] Starting HTTP server on endpoint localhost:8474   host=localhost port=8474 version=git
```

The basic verification that it is working:

```shell
$ toxiproxy-cli list
```

api.go
```diff
@@ -81,7 +81,7 @@ func (server *ApiServer) Listen(host string, port string) {
 		"host":    host,
 		"port":    port,
 		"version": Version,
-	}).Info("API HTTP server starting")
+	}).Printf("Toxiproxy HTTP server starting on endpoint %s:%s", host, port)
```
miry, line 84: It could be used [`Infof`](https://github.com/sirupsen/logrus/blob/79c5ab66aa2ce7d9ff7b3c437ebc22fcc519a967/entry.go#L361), so it would be more explicit which level is used.

Also I try avoid using the application name in the logs like `Toxiproxy`. It is already in the name of the binary `toxiproxy-server`. Otherwise it would overloaded with Toxiproxy word in logs, that most of time ignored.

miry:
> Sounds good, here's the updated output. Do you think this closes the linked issue? Or should the documentation be updated.

It is fine just to close the linked issue. The endpoint to check if toxiproxy ready is presented in README.

areveny:
```
dist % ./toxiproxy-server
INFO[0000] Starting HTTP server on endpoint localhost:8474  host=localhost port=8474 version=git
```

Thanks. It's now what you suggested. Really appreciate the feedback on a small change, I hope I can make more substantial contributions too.

miry (approved):

PR #359, opened 2022-01-19
Handle slicer Toxic with zero SizeVariation

### LINKED ISSUE #178: panic in slicer
Running on `2.1.1` with a `slicer` toxic:

```
~/P/shopify (master|✚1…) $ /Users/kir/Downloads/toxiproxy-server-darwin-amd64
INFO[0000] API HTTP server starting                      host=localhost port=8474 version=2.1.1
INFO[0053] Started proxy                                 name=dumb proxy=127.0.0.1:8182 upstream=localhost:8181
INFO[0177] Accepted client                               client=127.0.0.1:62635 name=dumb proxy=127.0.0.1:8182 upstream=localhost:8181
WARN[0177] Source terminated                             bytes=655582 err=read tcp 127.0.0.1:8182->127.0.0.1:62635: use of closed network connection name=dumb
INFO[0179] Accepted client                               client=127.0.0.1:62671 name=dumb proxy=127.0.0.1:8182 upstream=localhost:8181
WARN[0179] Source terminated                             bytes=655582 err=read tcp 127.0.0.1:8182->127.0.0.1:62671: use of closed network connection name=dumb
INFO[0184] Accepted client                               client=127.0.0.1:62739 name=dumb proxy=127.0.0.1:8182 upstream=localhost:8181
WARN[0184] Source terminated                             bytes=655582 err=read tcp 127.0.0.1:8182->127.0.0.1:62739: use of closed network connection name=dumb
INFO[0238] Accepted client                               client=127.0.0.1:63609 name=dumb proxy=127.0.0.1:8182 upstream=localhost:8181
WARN[0238] Source terminated                             bytes=655582 err=read tcp 127.0.0.1:8182->127.0.0.1:63609: use of closed network connection name=dumb
INFO[0271] Accepted client                               client=127.0.0.1:64135 name=dumb proxy=127.0.0.1:8182 upstream=localhost:8181
panic: invalid argument to Intn

goroutine 49 [running]:
panic(0x261fa0, 0xc420162050)
        /usr/local/Cellar/go/1.7.1/libexec/src/runtime/panic.go:500 +0x1a1
math/rand.(*Rand).Intn(0xc42008a080, 0x0, 0x2)
        /usr/local/Cellar/go/1.7.1/libexec/src/math/rand/rand.go:116 +0xd0
math/rand.Intn(0x0, 0x0)
        /usr/local/Cellar/go/1.7.1/libexec/src/math/rand/rand.go:247 +0x37
github.com/Shopify/toxiproxy/toxics.(*SlicerToxic).chunk(0xc4200eeb40, 0x0, 0x97, 0x1, 0x0, 0x0)
        /Users/jpittis/Code/go/src/github.com/Shopify/toxiproxy/toxics/slicer.go:42 +0xcc
github.com/Shopify/toxiproxy/toxics.(*SlicerToxic).Pipe(0xc4200eeb40, 0xc420180640)
        /Users/jpittis/Code/go/src/github.com/Shopify/toxiproxy/toxics/slicer.go:60 +0x139
github.com/Shopify/toxiproxy/toxics.(*ToxicStub).Run(0xc420180640, 0xc420074360)
        /Users/jpittis/Code/go/src/github.com/Shopify/toxiproxy/toxics/toxic.go:82 +0xb5
created by github.com/Shopify/toxiproxy.(*ToxicLink).Start
        /Users/jpittis/Code/go/src/github.com/Shopify/toxiproxy/link.go:73 +0x143
```

cc @jpittis

areveny:
This change skips randomized size variation for the slicer toxic if the `SizeVariation` parameter is 0, which is also the default setting.

Additionally, it fixes a small bug in the randomization. `rand.Intn(1)` was being added to `mid` to adjust for integer division rounding. However, `rand.Intn(n)` returns an integer in the range [0, n), so `rand.Intn(1)` always returns 0. It has been changed to `rand.Intn(2)`.

Closes #178


toxics/slicer_test.go
```diff
@@ -56,3 +56,51 @@ L:
...
+			case <-output:
+			}
+		}
+	}()
+
+	input <- &stream.StreamChunk{Data: data}
+
+	buf := make([]byte, 0, len(data))
+	reads := 0
+L:
```
miry, line 89: Is it possible to avoid Labels?

miry:
@areveny It looks like I still have problems random failing tests for Slicer.

```shell
$ MallocNanoZone=0 go test -v -race -timeout 1m ./...
...
=== RUN   TestSlicerToxic
    slicer_test.go:53: Expected to read about 480 times, but read 0 times.
    slicer_test.go:56: Server did not read correct buffer from client!
--- FAIL: TestSlicerToxic (0.02s)
```

areveny:
Reverted the timeout to its original duration.

miry:
Thank you @areveny. Can you please update the title of the current PR and entry in CHANGELOG to make it more informative. Example: Handle toxic slicer when SizeVariation is null.
It would remove duplication in history for readers like: Fix bug.

miry (approved):

Strech (approved):
LGTM 👍🏼 

I would suggest to take a look at the comment which might be outdated

toxics/slicer.go
```diff
@@ -37,10 +37,13 @@ func (t *SlicerToxic) chunk(start int, end int) []int {
 		return []int{start, end}
 	}
 
+	mid := start + (end-start)/2
 	// +1 in the size variation to offset favoring of smaller
 	// numbers by integer division
 	//#nosec
```
miry, line 43:
```suggestion
	//#nosec
	mid := start + (end-start)/2 + rand.Intn(2)

```

toxics/slicer.go
```diff
@@ -37,10 +37,13 @@ func (t *SlicerToxic) chunk(start int, end int) []int {
...
-	mid := start + (end-start)/2 + (rand.Intn(t.SizeVariation*2) - t.SizeVariation) + rand.Intn(1)
+	if t.SizeVariation > 0 {
+		mid = mid + (rand.Intn(t.SizeVariation*2) - t.SizeVariation) + rand.Intn(2)
```
miry, line 45:
```suggestion
		mid = mid + rand.Intn(t.SizeVariation*2) - t.SizeVariation
```

toxics/slicer.go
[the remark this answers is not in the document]
areveny: The comment is still valid. `rand.Intn(2)` is the correct call for creating a uniform random over {0, 1}. The comment is helpful because otherwise I could not have guessed what the call was for.

areveny: I propose keeping the comment. It is accurate to the code (creating a random on {0, 1}) and is helpful because otherwise I could not have guessed what the random integer was for.

toxics/slicer.go
areveny: I propose keeping the adjustment inside the `if`. Doing so has the nice property of making the slicing deterministic if `SizeVariation=0`.

With the code now, if someone has `SizeVariation=0`, they are guaranteed to receive their data in `NumberOfBytes / AverageSize` equally sized chunks (plus 1 if there is a remainder).

miry: The part `rand.Intn(2)` is connected to  to `(end-start)/2`. Other proposal is to remove `rand.Intn(2)`, as it was before behavior.  I think `rand.Intn(t.SizeVariation*2)` is enough for randomization.

toxics/slicer.go
miry: The comment suggested to always +1 and only related to part `(end-start)/2`. Currently the part of calculation size is bigger and comment did not cover all other parts. 

That's the reason we decided to remove it.

Other solution is to remove `rand.Intn(2)`

```suggestion
	mid := start + (end-start)/2
```

areveny:
I have gone with the approach of removing the `rand.Intn(2)`. The comment is removed as well.

Strech (approved):
Looks cleaner 👍🏼

neufeldtech:
Changes have been release as part of https://github.com/Shopify/toxiproxy/releases/tag/v2.4.0

Issue #365, opened 2022-02-25
Feature Request - Prometheus/OpenMetrics endpoint


neufeldtech:
It would be great to have a Prometheus or [OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md) endpoint in the HTTP api (`/metrics`) to be able to gain visibility on how much data is passing through a particular proxy.

A Common question that often comes up when using Toxiproxy in complex environments (with automation, in Kubernetes) is "is my data _actually_ flowing through Toxiproxy?". Having a standard metrics endpoint could help solve that.

Exposing these in a common metric format (Prometheus/OpenMetrics) rather than a custom format (arbitrary JSON or other) would enable wide adoption by folks that already have infrastructure deployed for ingesting Prometheus-style metrics.

For history, see also https://github.com/Shopify/toxiproxy/issues/266 where metrics were originally proposed (but through a custom format).

PR #366, opened 2022-02-25
Added prometheus metrics

neufeldtech:
Implements #365 

WIP


link.go
```diff
@@ -73,6 +78,12 @@ func (link *ToxicLink) Start(name string, source io.Reader, dest io.WriteCloser)
 				"err":   err,
 			}).Warn("Source terminated")
 		}
+		metrics.ReceivedBytesTotal.WithLabelValues(
+			streamDirection,
+			proxyName,
+			listenAddr,
+			upstreamAddr,
```
miry, line 85: @neufeldtech I prefer to share variables with arguments instead of closure. 
It would help to split method in future.

Abstract example:
```
go func(streamDirection, proxyName, ...) {
}(streamDirection, proxyName, ...)
```

link.go
```diff
@@ -111,6 +122,13 @@ func (link *ToxicLink) Start(name string, source io.Reader, dest io.WriteCloser)
 				"err":   err,
 			}).Warn("Destination terminated")
 		}
+
+		metrics.SentBytesTotal.WithLabelValues(
+			streamDirection,
+			proxyName,
+			listenAddr,
+			upstreamAddr,
+		).Add(float64(bytes))
```
miry, line 131: It would nice to have a context object with labels values. and then add metrics only in the code. it would remove duplication also partially related in the design to avoid using of global variables.

metrics/metrics.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+var (
+	proxyLabels = []string{
+		"direction",
+		"proxy",
+		"listener",
+		"upstream",
+	}
+	ReceivedBytesTotal = promauto.NewCounterVec(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
neufeldtech: I will look at refactoring this to be able to build the metric descriptions on demand via a function

link.go
neufeldtech: I'm not sure that this is possible in the Prometheus library (if you know how, please let me know). From how I've used these libraries, it's always `<metric> . <apply labels> . <apply values>` and I don't know that it's possible to do something like `<init labels> . <apply metric> . <apply values>`

link.go
neufeldtech: > It would help to have initialization part for metrics, than could be configured.

I'm glad you brought this up for discussion. I will refactor to make metrics optional (feature flagged) via a command line arg (and likely an environment variable as well, for easy Kubernetes compatibility).

neufeldtech:
We might need to discuss how to integrate this a little further - I just found another instance of NewProxy() where we don't have access to the server parent struct to get the metrics config data 

https://github.com/Shopify/toxiproxy/blob/master/proxy_collection.go#L97-L101

api.go
```diff
@@ -89,6 +100,28 @@ func (server *ApiServer) Listen(host string, port string) {
 	}
 }
 
+// metricsEnabled determines whether we have any prometheus metrics registered for exporting.
+func (server *ApiServer) metricsEnabled() bool {
+	if server.RuntimeMetricsEnabled || server.ProxyMetricsEnabled {
```
Eagerod, line 105: Are we opposed to straight up

```
return server.RuntimeMetricsEnabled || server.ProxyMetricsEnabled
```

metrics/common.go
```diff
@@ -0,0 +1,5 @@
+package metrics
+
+var (
```
Eagerod, line 3: const?

metrics/proxy.go
```diff
@@ -0,0 +1,46 @@
...
+	proxyLabels        []string
+	ReceivedBytesTotal *prometheus.CounterVec
+	SentBytesTotal     *prometheus.CounterVec
+}
+
+func (c *ProxyMetrics) Collectors() []prometheus.Collector {
+	return c.collectors
+}
+
+func NewProxyMetrics() *ProxyMetrics {
```
Eagerod, line 19: My Go Spidey sense tells me to basically never use pointers for structs unless the struct itself is pretty huge.

This struct is just 4 pointers on its own, so not "pretty huge". Also doesn't seem like it's heavily mutated itself, so probably doesn't need to have interfaces on the pointer itself. 

Maybe.

link.go
```diff
@@ -73,8 +77,11 @@ func (link *ToxicLink) Start(name string, source io.Reader, dest io.WriteCloser)
 				"err":   err,
 			}).Warn("Source terminated")
 		}
+		if server.Metrics.proxyMetricsEnabled() {
+			server.Metrics.ProxyMetrics.ReceivedBytesTotal.WithLabelValues(linkLabels...).Add(float64(bytes))
+		}
 		link.input.Close()
-	}()
+	}(labels)
```
miry, line 84: I see usage of `server.Metrics` inside goroutine. You can specify either `server` or `server.Metrics` in arguments.

metrics.go
```diff
@@ -0,0 +1,55 @@
...
+func (m *metricsContainer) proxyMetricsEnabled() bool {
+	return m.ProxyMetrics != nil
+}
+
+// anyMetricsEnabled determines whether we have any prometheus metrics registered for exporting.
+func (m *metricsContainer) anyMetricsEnabled() bool {
+	if m.runtimeMetricsEnabled() || m.proxyMetricsEnabled() {
+		return true
+	}
+	return false
```
miry, line 41: It is a good candidate to collapse all `if` statements to magic:

```go
func (m *metricsContainer) anyMetricsEnabled() bool {
  return m.runtimeMetricsEnabled() || m.proxyMetricsEnabled()
}
```

miry (approved):
Small comments left and linting errors. Overall it looks awesome.

metrics/proxy.go
miry: The method `NewStruct` usually returns pointer, otherwise it is a public struct that could be used without initialization method.

cmd/server/server.go
```diff
@@ -33,6 +37,10 @@ func parseArguments() cliArguments {
 		"JSON file containing proxies to create on startup")
 	flag.Int64Var(&result.seed, "seed", time.Now().UTC().UnixNano(),
 		"Seed for randomizing toxics with")
+	flag.BoolVar(&result.runtimeMetrics, "enable-runtime-metrics", false,
+		`enable runtime-related prometheus metrics (default "false")`)
+	flag.BoolVar(&result.proxyMetrics, "enable-proxy-metrics", false,
+		`enable toxiproxy-specific prometheus metrics (default "false")`)
```
neufeldtech, line 43: Should these flag names change to just `-runtime-metrics` and `-proxy-metrics`? The `enable` piece probably isn't necessary?

PR #375, opened 2022-03-07
Release 2.4.0

neufeldtech:
Note: had to update from centos to fedora to avoid rpm shenanigans with centos support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # [Unreleased]
 
+# [2.4.0] - 2022-03-07
+
 * Verify git tag on release (#347, @miry)
 * Fix MacOS 12 tests for go17 with -race flag (#351, @strech)
 * Rename `testing/` and `bin/` folders (#354, @strech)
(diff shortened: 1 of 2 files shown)


muz (approved):

miry:
> to avoid rpm shenanigans with centos support

@neufeldtech Can you share some links about centos and rpm support?

neufeldtech:
@miry I was getting an error when attempting to do a `yum install` because Centos tried to download yum cache (I don't have the error handy anymore). 

But anyway, this is some more context around Centos being end of life: https://www.zdnet.com/article/centos-linux-8-is-about-to-die-what-do-you-do-next/

> The end of [CentOS 8 Linux](https://www.centos.org/) has been coming for awhile now, and the day is finally here. On December 31, 2021, [Red Hat](https://www.redhat.com/en)'s[ CentOS Linux 8 will reach End Of Life (EOL)](https://www.centos.org/centos-linux-eol/). Since that falls right in the heart of the holiday season, Red Hat will extend CentOS Linux 8 zero-day support until January 31, 2022. Indeed, there will be one last CentOS Linux 8 release -- perhaps even after CentOS 8's official EOL. After that, it's all over for CentOS Linux.

miry:
Wow. Centos is dead!

PR #377, opened 2022-03-30
Basic Linux alternatives for chaotic tooling tc instead of nc

MUCZ:
I think there is a spelling error here, `nc` does not have the ability of network emulation, `tc`'s `netem` module does, and `tc/netem` module do require root privileges, which corresponds to the context

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ stopping you from creating a client in any other language (see
 ## Why yet another chaotic TCP proxy?
 
 The existing ones we found didn't provide the kind of dynamic API we needed for
-integration and unit testing. Linux tools like `nc` and so on are not
+integration and unit testing. Linux tools like `tc` and so on are not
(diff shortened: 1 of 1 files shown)


MUCZ:
Sorry, I've signed the CLA but I have no idea about how to make the bod re-check my CLA status.

miry:
@MUCZ I would think some `nc` hacks to simulate outages. Because `nc` could be used as proxy (some my hacks https://medium.com/jetthoughts/netcat-with-ssh-port-forwarding-148177b2e850). 

`tc` is more low level traffic shaper as I know. Could be also added.

I would think to remove this section at all.

miry:
> Sorry, I've signed the CLA but I have no idea about how to make the bot re-check my CLA status.

It is fine. I have re run the checker. Thanks for signing.

PR #392, opened 2022-06-01
Added Http Header toxic to use on Webservices deployed on Kubernetes

alfredosc8:
Http Header toxic to use on Webservices deployed on Kubernetes and an example of config.json to better configuration


alfredosc8:
Not required

PR #431, opened 2022-09-03
Update test-e2e script

miry:
Extract execution of benchmark script to function.
Use alias for command lines.
Extract trap eval string to function.

### DIFF
--- a/scripts/test-e2e
+++ b/scripts/test-e2e
@@ -2,36 +2,56 @@
 
 set -ueo pipefail
 
+cd "$(dirname "$0")"
+
+server="../dist/toxiproxy-server"
+state="started"
+
+benchmark() {
+  go test -bench=.. ../test/e2e -v
+}
+
+cli() {
+  ../dist/toxiproxy-cli "$@" 2>&1 | sed -e 's/^/[client] /'
+}
+
 wait_for_url() {
   curl -s --retry-connrefused --retry 5 --retry-delay 2 --retry-max-time 30 \
        --max-time 1 -L -I -X GET "${1}"
 }
 
 # Stop all background jobs on exit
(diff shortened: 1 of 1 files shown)


scripts/test-e2e
```diff
@@ -2,36 +2,58 @@
...
+cd $(dirname "$0")
+
+server="../dist/toxiproxy-server"
+
+benchmark() {
+  go test -bench=.. ../test/e2e -v
+}
+
+cli() {
+  ../dist/toxiproxy-cli $* 2>&1 | sed -e 's/^/[client] /'
```
barthalion, line 14:
```suggestion
  ../dist/toxiproxy-cli "$@" 2>&1 | sed -e 's/^/[client] /'
```

This is to ensure expanded argv is still quoted if arguments include special characters.

scripts/test-e2e
```diff
@@ -2,36 +2,58 @@
...
 wait_for_url() {
   curl -s --retry-connrefused --retry 5 --retry-delay 2 --retry-max-time 30 \
        --max-time 1 -L -I -X GET "${1}"
 }
 
 # Stop all background jobs on exit
-trap 'pkill -15 -f "dist/toxiproxy-server -proxy-metrics -runtime-metrics$"; pkill -15 -f "exe/endpoint$"' \
-     EXIT SIGINT SIGTERM
+function cleanup() {
+  sig="$1"
```
barthalion, line 24:
```suggestion
  local sig="$1"
```

Bash variables are global even if declared in a function.

barthalion (commented):
Just some nitpicks, feel free to ignore if they don't make sense in wider context.

scripts/test-e2e
miry: nice. fixed.

scripts/test-e2e
miry: thank you, fixed.

miry:
merged in commit 29685f4cd84df09161da9a58674484e9a83fcef2

PR #436, opened 2022-09-06
ToxicStub allow to unblocking write to Output

miry:
It could happen when Link has no reciever and there is some packets in buffer.
It produces deadlock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 * Show uniq request id in API HTTP response. (#425, @miry)
 * Add method to parse `stream.Direction` from string.
   Allow to convert `stream.Direction` to string. (#430, @miry)
+* Add posibility to write to Output with deadline.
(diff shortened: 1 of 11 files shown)


kwart:
This PR seems to cause incorrect HTTP responses on the `DELETE` command:

```bash
curl -v -X DELETE http://localhost:49266/proxies/tc-okJxHciQ:5701/toxics/CUT_CONNECTION_DOWNSTREAM
*   Trying 127.0.0.1:49266...
* Connected to localhost (127.0.0.1) port 49266 (#0)
> DELETE /proxies/tc-okJxHciQ:5701/toxics/CUT_CONNECTION_DOWNSTREAM HTTP/1.1
> Host: localhost:49266
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```

PR #441, opened 2022-09-11
Use single http client for cli

miry:
Introduce single Client httpClient for all cli requests.
It allows to add request headers in single place.
For example adds `User-Agent` HTTP header for all requests as `"toxiproxy-cli/<version> <os>/<runtime>"`.
Parse response on success or errors.
Set content type as json always.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -9,43 +9,56 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 )
 
 // Client holds information about where to connect to Toxiproxy.
 type Client struct {
-	endpoint string
+	UserAgent string
+	endpoint  string
+	http      *http.Client
 }
 
 // NewClient creates a new client which provides the base of all communication
 // with Toxiproxy. Endpoint is the address to the proxy (e.g. localhost:8474 if
 // not overridden).
 func NewClient(endpoint string) *Client {
-	if !strings.HasPrefix(endpoint, "https://") && !strings.HasPrefix(endpoint, "http://") {
+	if !strings.HasPrefix(endpoint, "https://") &&
+		!strings.HasPrefix(endpoint, "http://") {
 		endpoint = "http://" + endpoint
 	}
-	return &Client{endpoint: endpoint}
+
+	http := &http.Client{
+		Timeout: 30 * time.Second,
+	}
+
+	return &Client{
+		UserAgent: "toxiproxy-cli",
+		endpoint:  endpoint,
+		http:      http,
+	}
+}
+
+// Version returns a Toxiproxy running version.
+func (client *Client) Version() ([]byte, error) {
+	return client.get("/version")
 }
 
 // Proxies returns a map with all the proxies and their toxics.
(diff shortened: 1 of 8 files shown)


client/proxy.go
```diff
@@ -76,34 +68,22 @@ func (proxy *Proxy) Disable() error {
 // the proxy such as listen port and active toxics will be deleted as well. If you just wish to
 // stop and later enable a proxy, use `Enable()` and `Disable()`.
 func (proxy *Proxy) Delete() error {
-	httpClient := &http.Client{}
-	req, err := http.NewRequest("DELETE", proxy.client.endpoint+"/proxies/"+proxy.Name, nil)
-	if err != nil {
-		return err
+	err := proxy.client.delete("/proxies/" + proxy.Name)
+	if err == nil {
```
alileza, line 72: for the sake of consistency wouldnt it be better to 

```
if err != nil {
  return fmt.Errorf("Delete: %w", err)
}
```

then on line 75, just `return nil`?

alileza (approved):

client/client.go
```diff
@@ -213,10 +208,123 @@ func (client *Client) RemoveToxic(options *ToxicOptions) error {
...
+		return nil, err
+	}
+
+	resp, err := c.send(req, http.StatusOK)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	result, err := io.ReadAll(resp.Body)
```
alileza, line 267: I wonder if this should be part of `send` instead, since it's duplicated in 3 other function

client/client.go
```diff
@@ -213,10 +208,123 @@ func (client *Client) RemoveToxic(options *ToxicOptions) error {
...
+	err = c.validateResponse(resp, success_code)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (c *Client) validateResponse(resp *http.Response, expectedCode int) error {
+	if resp.StatusCode < 300 {
```
alileza, line 309: to indicate success, I think `if resp.StatusCode >= http.StatusOK && resp.StatusCode < http.Status StatusMultipleChoices` would be more precise, to exclude `1xx`, not sure if that edge cases would ever happen

client/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+		if ua != expected {
+			t.Errorf("User-Agent for %s %s is expected `%s', got: `%s'",
+				r.Method,
+				r.URL,
+				expected,
+				ua)
+		}
+
+		content_type := r.Header.Get("Content-Type")
```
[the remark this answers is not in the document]
miry: My bad behavior. Fixed.

client/proxy.go
miry: make sense. Fixed.

client/client.go
miry: I think it is too much. Toxiproxy does not have any 1xx response codes.
Updated the validation though.

I would still use numbers 200 and 300, instead of constants - just for simple readability.

client/client.go
miry: good catch.

PR #451, opened 2022-10-09
feat: add `corrupt` toxic

thalesmg:
Adds a toxic that flips random bits of the input stream, corrupting it.

Example run:

```
ͳ curl -v http://localhost:4001/api/echo -H "Content-Type: application/json" -d '{"echo": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}'
*   Trying 127.0.0.1:4001...
* Connected to localhost (127.0.0.1) port 4001 (#0)
> POST /api/echo HTTP/1.1
> Host: localhost:4001
> User-Agent: curl/7.85.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 77
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< cache-control: max-age=0, private, must-revalidate
< content-length: 76
< content-type: application/json; charset=utf-8
< date: Sun, 09 Oct 2022 17:55:36 GMT
< server: Cowboy
< x-request-id: Fxx4NnPZWF6J4lwAAABG
<
* Connection #0 to host localhost left intact
{"echo":"aaqaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}
```

(Note that it changed one of the `a`'s to a `q`.  Many times it may corrupt the request such that a response is not sent back.)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -40,3 +40,10 @@ linters-settings:
 
   misspell:
     locale: US
+issues:
+  exclude-rules:
+    - path: toxics/corrupt\.go
+      linters:
+        - gosec
+      # we don't need cryptographically secure RNGs for this
+      text: "G404:"
--- a/README.md
+++ b/README.md
@@ -446,6 +446,12 @@ Closes connection when transmitted data exceeded limit.
 
  - `bytes`: number of bytes it should transmit before connection is closed
 
+#### corrupt
+
+Flips random bits of the input stream, corrupting it.
+
+ - `probability`: probability of any given bit in the input of being flipped
+
 ### HTTP API
 
 All communication with the Toxiproxy daemon from the client happens through the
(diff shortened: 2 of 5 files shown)


thalesmg:
@miry thanks for the pointers. I'll look at them and update the PR.

miry:
@thalesmg I think it is a valid example. 

I agree, it would not be useful for most of users.
In same time it is nice to have a fleet of toxics 
to simulate not just TCP, but Application layer protocols.

As alternative: it is still possible to build a custom toxiproxy server with custom toxics.

I would still think the better place for new toxics in the that repo.

Issue #123, opened 2016-08-23
Simulate slow connection open


kevinhughes27:
I'm trying to use Toxiproxy to test a connection that is slow to open (I am specifically trying to test our Elastic Search configuration which has both a request_timeout and a request open_timeout which is shorter https://github.com/Shopify/shopify-app-store/pull/1297). 

My test fails after the full timeout, not the open timeout (4 vs 2 seconds). From a discussion with @Sirupsen he pointed out that this is probably because when the Elastic Search client is routed through Toxiproxy the Toxiproxy server accepts the connection immediately and then applies the toxic. From the Elastic Search client's perspective, it passed the request open_timeout for this reason.

For discussion
@Sirupsen @xthexder @jpittis

xthexder:
Unfortunately the way toxiproxy and TCP in general is set up, there's not a whole lot that can be done about this. If I remember correctly, when I tested this before, the kernel will accept / open the connection regardless of if `listener.Accept()` is called by the user.

Someone else might have some other ideas?

kevinhughes27:
I kind of figured this might not be possible to do..

Sirupsen:
If Toxiproxy because a L2 proxy it may be possible to delay the TCP opening handshake?

xthexder:
This was brought up once before: https://github.com/Shopify/toxiproxy/issues/72

xthexder:
Going to close this, since it's not really feasible and would pretty much require a rewrite. Maybe in the distant future.

PR #536, opened 2023-10-05
Introduce SlowOpenToxic to approximate the TCP handshake latency

stribioli:
As discussed in #123, the TCP handshake is performed by the kernel, so toxiproxy cannot inject latency there. Therefore, effects of latency on connection timeouts cannot be tested using this tool.

However, a user might be interested in the overall effects of latency from an application point of view. That is, the end-to-end effect on an entire conversation (TCP connection, application request, application response).

Using the `LatencyToxic` only, this end-to-end effect is not accurately modeled, because the initial RTT would be missing.
This pull request introduces a new toxic, `SlowOpenToxic`, that simulates the handshake latency by adding a delay to the first data packet only. I hope you'll find it useful!

### Latencies breakdown
In reality, you would have:
* a 1-RTT delay on the TCP handshake
* a 1-RTT delay on all subsequent data packets

Using Toxiproxy, with a `SlowOpenToxic` and a `LatencyToxic` (with the same delay, equal to 1 RTT), you would have:
* a 0-RTT delay on the TCP handshake
* a 2-RTT delay on the first data packet (1-RTT delay added by `SlowOpenToxic` and 1-RTT delay added by `LatencyToxic`)
* a 1-RTT delay on all subsequent data packets (added by `LatencyToxic`).

### Implementation notes
* The new toxic is derived from `LatencyToxic`. Once the first packet has been processed, a persisted State records the event and makes the toxic _warm_. From that point, this toxic is transparent.
* I extracted a block from the `LatencyToxic` tests (`doLatencyRound`) in order to:
  * Re-use it in the `SlowOpenToxic` tests.
  * Run it multiple times to test the different behavior between `LatencyToxic` and `SlowOpenToxic`. To do this, I also tweaked the harness (at toxic_test.go:69) to process multiple packets, instead of stopping at the first one.

### DIFF
--- a/scripts/test-e2e
+++ b/scripts/test-e2e
@@ -113,6 +113,28 @@ cli toxic delete --toxicName="latency_upstream" shopify_http
 
 echo -e "-----------------\n"
 
+echo "=== SlowOpen toxic downstream"
+
+cli toxic add --downstream \
+              --type=slow_open \
+              --toxicName="slow_open_downstream" \
+              --attribute="delay=1000" \
+              --toxicity=0.99 \
+              shopify_http
+cli inspect shopify_http
+
+benchmark
+
+cli toxic update --toxicName="slow_open_downstream" \
+                 --attribute="delay=500" \
+                 --toxicity=0.7 \
+                 shopify_http
+cli inspect shopify_http
+
+cli toxic delete --toxicName="slow_open_downstream" shopify_http
+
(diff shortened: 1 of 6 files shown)


BudaNET:
> THe workaround can be sufficient for my need. I need this so badly :)

Well I will probably build it locally.

PR #538, opened 2023-10-14
Fix invalid JSON in /version endpoint response

whatyouhide:
The `/version` endpoint returns invalid JSON, because the `\n` character at the end of the response is escaped by the backticks. This PR should fix that. I’m no Go expert, so I’m just winging this 🙃

### DIFF
--- a/api.go
+++ b/api.go
@@ -477,7 +477,7 @@ func (server *ApiServer) Version(response http.ResponseWriter, request *http.Req
 	log := zerolog.Ctx(request.Context())
 
 	response.Header().Set("Content-Type", "application/json;charset=utf-8")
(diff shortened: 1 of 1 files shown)


whatyouhide:
I have signed the CLA!

miry:
cc: @abecevello

whatyouhide:
Ah, yes. Sorry, I didn't even run tests locally cause I don't work in Go and had nothing set up for it. I took the laziest approach 😅 

Anyways, I pushed a commit that should fix the test 🤞

abecevello (approved):

abecevello:
@whatyouhide thanks for the fix. I've merged it and I'll see about doing a release that includes this soon.

PR #604, opened 2024-12-09
docs(client): make example in README compile

samlaf:


client/README.md
```diff
@@ -114,6 +114,10 @@ func init() {
     if err != nil {
         panic(err)
     }
+   	proxies, err = toxiClient.Proxies()
+   	if err != nil {
+        panic(err)
+   	}
```
abecevello, line 120: This addition seems incorrect as `proxies` is assigned on line 108 as a result of the `toxiClient.Populate` call.

If this is meant as another example, similar to the example on line 122, I would be okay with this remaining but the indentation should be fixed, it should be commented out, and a comment added explaining what it does if its uncommented.

abecevello (commented):
Thanks for the contribution! I had a question and potential suggestion.

Also, in order to merge this change, the CLA must be signed using https://cla.shopify.com/ and then comment on this PR with "I have signed the CLA!" to re-run the checks.

abecevello:
Thank you for the PR and bringing this to our attention. Actually fixing this example to compile with the current API required some more involved changes. I've made https://github.com/Shopify/toxiproxy/pull/607 which does all the updates. I'm going to close this PR in favour of that one.

PR #607, opened 2024-12-11
Update client readme example to compile with previous API changes

abecevello:
Updates the full example in the client readme to compile. It was broken and didn't reflect the current API. I also removed an unnecessary example test that wasn't necessary to get a feel for how to use Toxiproxy.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -96,16 +96,15 @@ import (
     "time"
 
     toxiproxy "github.com/Shopify/toxiproxy/v2/client"
-    "github.com/garyburd/redigo/redis"
+    "github.com/gomodule/redigo/redis"
 )
 
 var toxiClient *toxiproxy.Client
-var proxies map[string]*toxiproxy.Proxy
 
 func init() {
     var err error
     toxiClient = toxiproxy.NewClient("localhost:8474")
-    proxies, err = toxiClient.Populate([]toxiproxy.Proxy{{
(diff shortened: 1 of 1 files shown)


client/README.md
```diff
@@ -96,16 +96,15 @@ import (
     "time"
```
mattbardal, line 96: Small typo that can be fixed on line 23 [here](https://github.com/Shopify/toxiproxy/pull/607/files#diff-c4577d0b3a9c55c4e75639bb45f3dd659e83bc31b7a30943fda6ef84f7cef47aR23)

client/README.md
```diff
@@ -149,17 +150,4 @@ func TestRedisBackendSlow(t *testing.T) {
...
-    defer proxy.Delete()
-
-    // Test connection through proxy.Listen
-    resp, err := http.Get("http://" + proxy.Listen)
-    if err != nil {
-        t.Fatal(err)
-    } else if resp.StatusCode != 200 {
-        t.Fatal("Proxy to google failed:", resp.StatusCode)
-    }
-}
```
mattbardal, line 164: Since we're planning on cleaning up this test, we could probably get rid of the `net/http` import as well since this is the only test that uses it.

mattbardal (approved):
Left a couple of small suggestions but otherwise this looks good to me!

PR #756, opened 2026-07-26
docs: clarify that latency toxics are one-way

### LINKED ISSUE #655: Latency toxic is not bidirectional
I'm using toxiproxy to test an networked game that uses [Cristian's Algorithm](https://en.wikipedia.org/wiki/Cristian's_algorithm) to synchronize a client clock with a server clock.  On a typical TCP connection, latency will be symmetric after the first few messages are exchanged.

Based on my testing, the `latency` toxic is unidirectional, applying only to host-to-client communications.  This asymmetric delay breaks the assumptions underlying this clock sync method.  When running client and server on the same PC with a toxiproxy between them, a clock sync request dispatched at T=475218ms reaches the server at T=475218ms (ie, immediately) and the reply is received at T=475278ms (with the 60ms latency I configured).

```
TRACE  time sync_receive: Roundtrip 60ms (client real time 180475218~180475278)
        Host's real time: 180475218
```

Furthermore, the unidirectional latency is undocumented:  indeed, the README implies otherwise:

> **latency**
> Add a delay to all data going through the proxy. The delay is equal to `latency` +/- `jitter`.

If latency were indeed applied to all data going through the proxy, I would expect a `latency` toxic valued at 25ms to produce 50ms of roundtrip delay.  But this is not the case (and this project seems much too mature to change the default behavior).

So, I recommend updating the documentation to reflect the "one-way" nature of the latency toxic.

In the future, a bidirectional mode for the latency toxic would be very useful for applications like mine.  Real-world LAN/intranet/internet applications are infinitely more likely to encounter symmetric latency than

Hashim1999164:
## Summary
- The latency toxic docs said the delay applied to all data through the proxy, which read like a bidirectional round trip.
- Latency toxics only affect the stream they are attached to (default `downstream`). Symmetric delay needs latency on both `upstream` and `downstream`.
- Updates the latency section to match the existing stream field behavior.

Fixes #655

## Test plan
- [ ] Read the latency section in README and confirm it matches the `stream` field docs
- [ ] No code changes; docs only

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,12 @@ For documentation on implementing custom toxics, see [CREATING_TOXICS.md](./CREA
 
 #### latency
 
-Add a delay to all data going through the proxy. The delay is equal to `latency` +/- `jitter`.
+Add a delay to data on a single proxy stream. The delay is equal to `latency` +/- `jitter`.
+
+Toxics are directional: by default a latency toxic is attached to the `downstream`
+stream (`server -> client`). It does not delay the opposite direction unless you
+also add a latency toxic with `stream=upstream`. For a symmetric round-trip delay,
+add latency on both streams.
 
 Attributes:
 


Hashim1999164:
I have signed the CLA!

Hashim1999164:
I have signed the CLA!

PR #764, opened 2026-08-25
fix(toxics/bandwidth): don't hang or panic when rate <= 0

jaideeppyne:
### Problem

`BandwidthToxic.Pipe` already treats a non-positive `rate` as "no throttling" (`if t.Rate <= 0 { sleep = 0 }`), but the packet-splitting loop used `t.Rate*100` as the split size unconditionally:

- **`rate == 0`** → split size is `0`, so `len(p.Data) > 0` stays true forever; the loop spins emitting empty chunks (`p.Data[:0]`) and never forwards the payload — the connection hangs.
- **`rate < 0`** → split size is negative; `p.Data[:negative]` panics with `slice bounds out of range`.

A bandwidth toxic created with `rate: 0` (the zero value when the field is omitted) is enough to trigger the hang; there is no input validation rejecting `rate <= 0`.

### Fix

Guard the split loop on `t.Rate > 0` so a non-positive rate falls through to the normal unthrottled send path — matching the existing `t.Rate <= 0` intent.

### Tests

Added `TestBandwidthToxicNonPositiveRate` (`rate_0`, `rate_-1`), which fails before the change (empty forward / panic) and passes after. Full `toxics` package passes; `gofmt`/`go vet` clean.

### DIFF
--- a/toxics/bandwidth.go
+++ b/toxics/bandwidth.go
@@ -38,8 +38,11 @@ func (t *BandwidthToxic) Pipe(stub *ToxicStub) {
 			} else {
 				sleep += time.Duration(len(p.Data)) * time.Millisecond / time.Duration(t.Rate)
 			}
-			// If the rate is low enough, split the packet up and send in 100 millisecond intervals
-			for int64(len(p.Data)) > t.Rate*100 {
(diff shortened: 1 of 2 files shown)


jaideeppyne:
I've signed the CLA — could the check be re-run?

PR #770, opened 2026-09-07
fix(toxics/slicer): never overflow the stack or slice out of range

### LINKED ISSUE #541: panic: runtime error: slice bounds out of range [808:773]
This panic happens when using the `slicer` toxic:

```
panic: runtime error: slice bounds out of range [808:773]

goroutine 114 [running]:
github.com/Shopify/toxiproxy/v2/toxics.(*SlicerToxic).Pipe(0x1400020c0f0, 0x1400021e480)
	github.com/Shopify/toxiproxy/v2/toxics/slicer.go:65 +0x320
github.com/Shopify/toxiproxy/v2/toxics.(*ToxicStub).Run(0x1400021e480, 0x14000210d20)
	github.com/Shopify/toxiproxy/v2/toxics/toxic.go:86 +0xb8
created by github.com/Shopify/toxiproxy/v2.(*ToxicLink).Start in goroutine 58
	github.com/Shopify/toxiproxy/v2/link.go:109 +0x2c8
```

The bug is here:
https://github.com/Shopify/toxiproxy/blob/7abcb6f9cf95d2fb27244435e22b16bda14ab5c3/toxics/slicer.go#L42-L44

`rand.Intn` can return a number in `[0, SizeVariation*2)`, so it can be less than `SizeVariation`, resulting a negative RHS and a `mid` less than `start`

lenamonj:
Fixes #541.

`SlicerToxic.chunk` recurses on an interval that never shrinks when `average_size` is 0, which is what `toxiproxy-cli toxic add -t slicer <proxy>` sends when no attributes are given: the first byte through the proxy ends the server with `fatal error: stack overflow`. With `size_variation` at or above `average_size` the jittered midpoint can land outside the piece, and `Pipe` panics with `slice bounds out of range` (#541).

`chunk` now returns the piece whole when `average_size` is 0 or less or the piece is under two bytes, and clamps the midpoint inside the piece so both halves are shorter than it. Slicing for attributes that already worked is unchanged.

The new test drives nine attribute pairs, including the zero defaults and `size_variation > average_size`, over six payload sizes and checks the pieces reassemble byte for byte. On main it dies with the stack overflow.

`make test`, `go vet ./...` pass.

### DIFF
--- a/toxics/slicer.go
+++ b/toxics/slicer.go
@@ -33,7 +33,9 @@ func (t *SlicerToxic) chunk(start int, end int) []int {
 	// If the size is within the random varation, _or already
 	// less than the average size_, just return it.
 	// Otherwise split the chunk in about two, and recurse.
-	if (end-start)-t.AverageSize <= t.SizeVariation {
+	// An average size of zero or less slices nothing, and a piece of
+	// fewer than two bytes cannot be split.
+	if t.AverageSize <= 0 || end-start < 2 || (end-start)-t.AverageSize <= t.SizeVariation {
 		return []int{start, end}
 	}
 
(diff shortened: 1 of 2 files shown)


lenamonj:
I have signed the CLA!

